PR #23, opened 2020-01-13
Added Enterprise as a difftools path

chrissie1:
Added Enterprise as another path to the devenv.exe file for the difftools. And updated the docs.


SimonCropp:
thanks

SimonCropp:

PR #137, opened 2020-05-27
add inline guid processing

HEskandari:
Scrubs the guid values within a string

### DIFF
--- a/src/Verify.Tests/Serialization/SerializationTests.ShouldScrubInlineGuidsByDefault.verified.txt
+++ b/src/Verify.Tests/Serialization/SerializationTests.ShouldScrubInlineGuidsByDefault.verified.txt
@@ -1,6 +1,6 @@
 {
-  Title: item Guid_1 - (ID={Guid_1}),
+  Title: 'item Guid_1 - (ID={Guid_1})',
   Variant: {
-    Id: variant id: Guid_2
+    Id: 'variant id: Guid_2'
   }
 }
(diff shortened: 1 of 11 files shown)


SimonCropp:
can u pull in the upstream. you are a few commits behind

HEskandari:
@SimonCropp done

src/Verify.Tests/Serialization/SerializationTests.ShouldScrubInlineGuidsByDefault.verified.txt
```diff
@@ -0,0 +1,6 @@
+{
+  Title: item Guid_1 - (ID={Guid_1}),
+  Variant: {
+    Id: variant id: Guid_2
```
SimonCropp, line 4: why have the single quotes around the text gone?

SimonCropp:
can you add some doco for this feature

src/Verify.Tests/Serialization/SerializationTests.ShouldScrubInlineGuidsByDefault.verified.txt
HEskandari: Good catch. Thing is that up the stack there is a code that does this:

```
if (sharedScrubber.TryConvertString(valueAsString, out var result))
{
    writer.WriteRawValue(result);
    return;
}
```

The `WriteRawValues` expects you to have quotes in place. Changing it to `WriteValues` puts the quotes for all the values. I'll send an update that'd put in the quotes manually when formatting the string.

HEskandari:
@SimonCropp added some snippets.

SimonCropp:

note i made it an opt in feature for backwards compat reasons

PR #139, opened 2020-05-30
Tweak commands copied to clipboard to work with both cmd.exe and powershell on Windows

gbiellem:
Currently the move and del comand copied to the clipboard are specific to cmd.exe  and do no work if pasted into a powershell console.  This change addresses this by pre-pending  `cmd /c ` to the front of the commands.  The result is that the commands work in both cmd and powershell


SimonCropp:

PR #312, opened 2021-03-10
Use ExtensionOrTxt in InnerVerifier so a possible extension override in the settings won't go unnoticed.

### LINKED ISSUE #307: WIP: Extension setting is being ignored.
I'll be investigating this further next week, but only so much for the time being:
In Verify 10.9.1, the extension setting seems to be ignored. Although I set it to "json", Verify fails with *received.txt" files.

Verification is being called like this:
```
settings.UseExtension("json");

return Verifier.Verify(element, settings);
```

I could debug it as far as you can see on the pictures below. As I wrote, I'll put more effort into it after the weekend.

image
image

danielcweber:
Fixes #307.

### DIFF
--- a/src/Verify/Verifier/InnerVerifier_Inner.cs
+++ b/src/Verify/Verifier/InnerVerifier_Inner.cs
@@ -82,7 +82,7 @@ namespace VerifyTests
                 return true;
             }
 
-            extension = "txt";
+            extension = settings.ExtensionOrTxt();
 
             if (VerifierSettings.StrictJson)
             {


SimonCropp:
i dont think this is correct. i thought about this some more, and i dont think overriding the serialization extension should be supported. it defaults to text, and if people want json they can use UseStrictJson. which is waht i did here https://github.com/VerifyTests/Verify/commit/17744d707cc8ad2245ebab385cf3e4ddcf01478b 

thoughts?

danielcweber:
I have no real opinion about the extension policy. I'm good with UseStrictJson. Maybe there shouldn't even be UseStrictJson because the naming doesn't really indicate to me that it's got something to do with file extensions.

But if UseExtension has a use case (like streams), what'll UseStrictJson do then? It seems a bit convoluted to have multiple options that may even be ignored sometimes.

Maybe there shouldn't be any possibility to override the extension at all.

danielcweber:
I'd also happily change all my *verified* files to txt if that's the new policy. I just figured updating Verify broke my snapshots so it seemed like a bug, but from looking at the sources, I already got the idea that this might actually be a new policy.

Issue #396, opened 2021-09-02
Property order in files written by Verify is machine-dependent (and I can't find a way to sort them)


bmueller84:
I use Verify and Verify.Xunit (11.24.0 on Framework 4.8) and have problems with inconsistent property orderings in the generated text files, which is why my tests will fail on some machines (in some sense, similar to [this bug](https://github.com/VerifyTests/Verify/issues/385)). I don't really know _why_ the ordering differs across machines, though...

Anyway, is there a known workaround for this (except for slapping json tags onto the objects to be serialized)? My initial approach was to write a custom Json contract resolver

    public class OrderedContractResolver : DefaultContractResolver
    {
        protected override IList<JsonProperty> CreateProperties(Type type, MemberSerialization memberSerialization)
            => base.CreateProperties(type, memberSerialization)
                .OrderBy(p => p.Order ?? int.MaxValue)
                .ThenBy(p => p.PropertyName)
                .ToList();
    }

and to pass it to the serializer:

    return Verifier.Verify(objectUnderTest).AddExtraSettings(s => s.ContractResolver = new OrderedContractResolver());

This kinda works but breaks a lot of things down the road. For example,

    return Verifier.Verify(objectUnderTest)
        .AddExtraSettings(s => s.ContractResolver = new OrderedContractResolver())
        .ModifySerialization(s => s.IgnoreMember<SomeType>(p => p.SomeProperty))

won't work anymore because, as far as I understand, Verify makes use of the (non-public) `CustomContractResolver` to take care of things like ignored properties. If `CustomContractResolver` was public, I might simply be able to derive my `OrderedContractResolver` from it. On the other hand, it feels like having a consistent property order should be general goal (and not something the user has to implement). Any thoughts on this?

SimonCropp:
firstly. this is a bug and you should not need to work around it.

Can you share the model, and the different orders it produces?

Do the machines have different .net sdks installed?

bmueller84:
Still working on a minimal reproduction (code under test has quite heavy dependencies and my initial attempt with just deleting 90% of the code failed. But I'm working on it). Both machines run .Net 4.8, but one of them is Windows 10 and the other Windows Server 2019 (shouldn't matter, I guess, just mentioning because the frameworks have [different release numbers](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed)).

[...]

PR #398, opened 2021-09-09
Allow user to sort serialized properties alphabetically

bmueller84:
As discussed in [this issue](https://github.com/VerifyTests/Verify/issues/396#issuecomment-915913731), I think it's useful to allow the user to enforce a consistent, alphabetical ordering of properties in the serialized data. A simple scenario that comes to mind is reordering members within a class, which would currently cause all corresponding tests to fail.

I've started with the most simple way to implement this (including a few tests). Please let me know what's missing.

### DIFF
--- a/src/Verify.Tests/Serialization/SerializationTests.SortingPropertiesAlphabeticallyDoesNotAffectTypeName.verified.txt
+++ b/src/Verify.Tests/Serialization/SerializationTests.SortingPropertiesAlphabeticallyDoesNotAffectTypeName.verified.txt
@@ -1,0 +1,19 @@
+﻿{
+  $type: Person,
+  Address: {
+    $type: Address,
+    Country: USA,
+    Street: 1 Puddle Lane
+  },
+  Children: {
+    $type: List<string>,
+    $values: [
+      Sam,
+      Mary
+    ]
+  },
+  FamilyName: Smith,
+  GivenNames: John,
+  Id: Guid_1,
+  Spouse: Jill
+}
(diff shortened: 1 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=398) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=398) before we can accept your contribution.<br/><br/>:x: bmueller84 [sign now](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=398)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/VerifyTests/Verify?pullRequest=398) it.</sub>

SimonCropp:
i was assuming you would want this to be a single static config option?

bmueller84:
Well, for versions before C# 9, we don't have module initializers, so each test class would have to to set this static config option anyway. But I guess it's not a bad idea to support both use cases. What do you think?

SimonCropp:
while some cant use module initializers, you can still use a test startup in xunit/nunit/mstest. or u can use a static contractor in a test base class. i just dont see it bein required at the per-test level

bmueller84:
I don't like these static settings general because they often make things harder to test. But it's your library, so I'm not gonna argue with you ;) How do you like the updated pull request?

Btw: NamerTests.ThrowOnConflict seems to be flaky (even on the original repo)

PR #430, opened 2021-11-02
Improve the message when a prefix is used multiple times

0xced:
It is handy to have a hint at disabling the uniqueness validation with `VerifierSettings.DisableRequireUniquePrefix()` directly into the exception message.


SimonCropp:
yep. looks good

PR #470, opened 2022-02-03
Undo removal of 'Verifier.' in docs

Siphonophora:
Hey, looks like find replace in 41e5e4f086fe668e9b480a2ca553cfb9af02abc1 broke the example docs.

### DIFF
--- a/docs/anonymous-types.md
+++ b/docs/anonymous-types.md
@@ -29,7 +29,7 @@ public Task Anon()
         FamilyName = "Aguirre"
     };
 
-    return Verify(
+    return Verifier.Verify(
         new
         {
             person1,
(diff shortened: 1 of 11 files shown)


Siphonophora:
Oh, I just noticed all the markdown is being built from those snippets. So maybe there just needs to be a comment on the readme next to the one about implicit usings.

SimonCropp:
you mean this comment? https://github.com/VerifyTests/Verify#usage

> All examples use [ImplicitUsings](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#implicitusings). Ensure the following is set to have examples compile correctly <ImplicitUsings>enable</ImplicitUsings>

Siphonophora:
Yeah. A comment there that you are expecting the static using would make sense.

SimonCropp:
there is already a comment

Issue #456, opened 2022-01-13
Verify.Xunit fails when .sln file is located in '/' directory


OlegKarasik:
All Verify.Xunit tests are failing when run on a solution located in '/' directory with the following exception:

```
System.ArgumentException : Empty name is not legal. (Parameter 'oldValue')
   at System.Text.StringBuilder.Replace(String oldValue, String newValue, Int32 startIndex, Int32 count)
   at System.Text.StringBuilder.Replace(String oldValue, String newValue)
   at ApplyScrubbers.<>c__DisplayClass5_0.<GetReplacements>b__2(StringBuilder builder) in /_/src/Verify/Serialization/Scrubbers/ApplyScrubbers.cs:line 83
   at ApplyScrubbers.Apply(String extension, StringBuilder target, VerifySettings settings) in /_/src/Verify/Serialization/Scrubbers/ApplyScrubbers.cs:line 106
   at VerifyEngine.GetResult(VerifySettings settings, FilePair filePair, Target target, Boolean previousTextHasFailed) in /_/src/Verify/Verifier/VerifyEngine.cs:line 38
   at VerifyEngine.HandleResults(List`1 targetList) in /_/src/Verify/Verifier/VerifyEngine.cs:line 60
   at InnerVerifier.VerifyInner(Object target, Func`1 cleanup, IEnumerable`1 targets) in /_/src/Verify/Verifier/InnerVerifier_Inner.cs:line 23
   at InnerVerifier.Verify[T](T target) in /_/src/Verify/Verifier/InnerVerifier_Json.cs:line 73
   at VerifyXunit.Verifier.<>c__DisplayClass1_0.<<Verify>b__0>d.MoveNext() in /_/src/Verify.Xunit/Verifier.cs:line 32
```

The issue is easily reproduced when running tests from a `Dockerfile`. Here is a small example of such `Dockerfile`:

```
# -- build --
FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build

WORKDIR /

# copy *.sln
COPY /CSI.ASQ.Assessments.sln

# copy *.csproj
COPY /src/Core.Generators/*.csproj           /src/Core.Generators/
COPY /tests/Core.Generators.Tests/*.csproj   /tests/Core.Generators.Tests/

# dotnet restore
RUN dotnet restore

# copy source files
COPY /src/Core.Generators/src             /src/Core.Generators/src
COPY /tests/Core.Generators.Tests/src     /tests/Core.Generators.Tests/src

[...]

PR #479, opened 2022-02-18
Fixing failure when .sln is located in '/' root directory.

OlegKarasik:
This PR is a fix for #456.

@SimonCropp is there any way to verify unit-tests in some independent environment? 

By some reason I have different tests failing on different machines. I have Mac M1 and Windows machines and I get failed tests even from `main` branch (when running using `dotnet test` command). What is interesting all these test run fine if run using VS Code Codelense.

Besides the fix, I have added a reference `Xunit.SkippableFact` package to fence some tests which aren't run on Mac (at least on M1 with .NET ARM). 

If some of these changes seem inappropriate feel free to remove them (I was intended to do some good, not vice versa).


OlegKarasik:
Oh, I see a PR hook. Okay I will investigate and try to understand how to fix the failing tests.

OlegKarasik:
I have found a reason - the tests I have created caused a cascade failure because they where reinitialising the global replacements dictionary. I have commented my tests out because refactoring of `ApplyScrubbers.cs` is out of scope of the issue.

src/Verify/Serialization/Scrubbers/ApplyScrubbers.cs
```diff
@@ -65,9 +67,12 @@ static void AddProjectAndSolutionReplacements(string? solutionDir, string projec
         {
             return;
         }
+
+        var regex = new Regex(@"([^/\\])[/\\]+$");
```
SimonCropp, line 71: can u do this without a regex?

src/Verify.Tests/ApplyScrubbersTests.cs
```diff
@@ -0,0 +1,25 @@
+// public class ApplyScrubbersTests
```
SimonCropp, line 1: why is this commented out?

src/Verify/Serialization/Scrubbers/ApplyScrubbers.cs
OlegKarasik: I think so, what are the concerns related to regex?

SimonCropp:
> Besides the fix, I have added a reference Xunit.SkippableFact package to fence some tests which aren't run on Mac (at least on M1 with .NET ARM).

can those tests be modified to work on an M1?

src/Verify.Tests/ApplyScrubbersTests.cs
OlegKarasik: So, I should have deleted the file. The reason for this is that having these tests uncommented breaks other tests in random (because ApplyScrubbers is a static class with global state and changing it during tests break other tests).

src/Verify/Serialization/Scrubbers/ApplyScrubbers.cs
SimonCropp: regex is slower, uses more memory, and more difficult to read than standard string manipulation

src/Verify.Tests/ApplyScrubbersTests.cs
SimonCropp: move it to a different test project so it runs in isolation

src/Verify/Serialization/Scrubbers/ApplyScrubbers.cs
OlegKarasik: Okay, I will rewrite the logic to use common string manipulations.

Issue #482, opened 2022-02-23
are there global settings?


MihailsKuzmins:
Tried asking on stackoverflow, but to no avail (https://stackoverflow.com/questions/71210285/does-verifytests-verify-have-global-settings)

In the [docs](https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#directory) it is mentioned how to setup a custom output directory for `Verify` and it works fine, but it is pain to do it for every test.
```cs
await Verify("value")
   .UseDirectory("CustomDirectory");
```
ApprovalTests have a global `UseApprovalSubdirectory` which solves this issue. So I wonder is there anything similar in here?  
The reason is that VS2022 does not allow renaming nested files, so I need to go to the file explorer and rename files there (pain...)

SimonCropp:
you should be able to use [DerivePathInfo](https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#derivepathinfo) to achieve this. So in your case

```
VerifierSettings.DerivePathInfo(
    (sourceFile, projectDirectory, type, method) =>
    {
        return new(
            directory: Path.Combine(projectDirectory, "CustomDirectory"),
            typeName: type.Name,
            methodName: method.Name);
    });
```

MihailsKuzmins:
@SimonCropp, is there also `VerifierSettings.something` for `classLevelSettings.UseExtension("json");`? I cannot seem to find it :(

SimonCropp:
is this what you are after https://github.com/VerifyTests/Verify/blob/main/docs/serializer-settings.md#usestrictjson  ?

MihailsKuzmins:
hmm, looks like, but it does not seem to work because the output file still has the `.txt` extension.

`DerivedPathInfo` works as expected and I get the output file where I need, but the extension is not changed to `.json`. (maybe I will try to checkout the source code and find out what goes wrong on my side)

![image](https://user-images.githubusercontent.com/47413092/155704343-b4a9fdb8-56f6-4f2e-a7b7-a10be3233cfd.png)
![image](https://user-images.githubusercontent.com/47413092/155704849-01fc8601-3a1e-4319-87ce-56121e48c5ed.png)

SimonCropp:
are you calling `UseStrictJson ` in a ModuleInitializer?

MihailsKuzmins:
[...]

PR #500, opened 2022-03-13
Failing test for .json extension for strings

MihailsKuzmins:
Issue #482

### DIFF
--- a/src/StrictJsonTests/Tests.JsonAsString.verified.txt
+++ b/src/StrictJsonTests/Tests.JsonAsString.verified.txt
@@ -1,0 +1,4 @@
+﻿{
+  "Prop1": "string value",
+  "Prop2": 123
+}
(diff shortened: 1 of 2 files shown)


SimonCropp:
so strict json only applies when you pass in an object to be serialized

```
        await Verify(new
        {
            Prop1 = "string value",
            Prop2 = 123
        });
```

when you pass in a string to `Verify()`it treats it as a string. this is necessary since if you do 

```
Verify("this is a string");
```
there is no way to distinguish between you case and a simple string.


If you have a string, and want to to be treated like json you need to call [VerifyJson](https://github.com/VerifyTests/Verify/blob/main/readme.md#verifyjson)

SimonCropp:
> configuring an extension globally which is not supported at the moment as I understand

correct since i dont see how that would be a good idea to have anything other than text.  passing in an arbitrary string, it would be pretty rare for any use to always want that to be non txt.

MihailsKuzmins:
then I guess we can close this PR and https://github.com/VerifyTests/Verify/issues/482 as well

PR #503, opened 2022-03-21
Source Generator - Hint Name

mrkwllce:
Extended the API to allow for passing along a hintName from Source Generators to allow the generator to use the existing unique name, so that the generated files will always be compared correctly.


SimonCropp:
> so that the generated files will always be compared correctly.

can you give me some scenarios where files are not correctly compared

mrkwllce:
Here's a branch that gives the example.

https://github.com/mrkwllce/Verify.SourceGenerators/tree/SampleBranch

SimonCropp:
i changes it so it sorts by hintname https://github.com/VerifyTests/Verify.SourceGenerators/pull/60

do this fix you issue?

SimonCropp:
can you try Verify.SourceGenerators 1.5.0-beta.1

PR #526, opened 2022-05-23
remove NumericId feature

SimonCropp:
Decided this was a bad idea since it often incorrectly conflates matching ids that should not be correlated


chrismcv:
Hi, 
Have just upgraded to this version, and found that all our snapshots are failing because of this change.  I can see from the diff that there is no way to opt back into the feature.

As we have `int` Id fields, I'm wondering what the recommendation for handling parallel integration tests is. (Where each arrange step is e.g. inserting into a Client table, and then snapshot an API response containing that Client record)

Thanks
Chris

SimonCropp:
> As we have int Id fields, I'm wondering what the recommendation for handling parallel integration tests is. (Where each arrange step is e.g. inserting into a Client table, and then snapshot an API response containing that Client record)

can u share a minimal repro as a test in a pull request

PR #548, opened 2022-06-27
Use CustomValueProvider for ShouldSerialize checks

wilversings:
By using valueProvider.GetValue directly we can encounter members that throw exceptions which were explicitly ignored by the user. Therefore we need to use the CustomValueProvider in order to handle such ignored exceptions.

### DIFF
--- a/src/Verify.Tests/Serialization/SerializationTests.cs
+++ b/src/Verify.Tests/Serialization/SerializationTests.cs
@@ -1941,11 +1941,11 @@ public class SerializationTests
 
     class WithCustomException
     {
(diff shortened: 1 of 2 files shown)


SimonCropp:
@wilversings looks good. any reason why it is still a draft?

wilversings:
Hi @SimonCropp 
I just wanted to check a few more things.
Is it necessary to create an Issue for this PR?

SimonCropp:
> Is it necessary to create an Issue for this PR?

nope.

SimonCropp:
thanks. this is now deployed

PR #601, opened 2022-08-16
Add static API support for MSTest

TomW-Skyline:
https://github.com/orgs/VerifyTests/discussions/587


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=601) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=601) before we can accept your contribution.<br/><br/>:x: TomW-Skyline [sign now](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=601)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/VerifyTests/Verify?pullRequest=601) it.</sub>

src/Verify.MSTest/Verifier.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        GetFileConvention fileConvention = (uniquenessForReceived, uniquenessForVerified) =>
+            ReflectionFileNameBuilder.FileNamePrefix(method, type, sourceFile, settings, uniquenessForReceived, uniquenessForVerified);
+
+        return new(sourceFile, settings, fileConvention);
+    }
+
+    static MethodInfo GetTestMethod()
+    {
+        var stackTrace = new StackTrace();
```
SimonCropp, line 21: stack walking is unreliable and very diffcult to debug. so this approach is not viable

TomW-Skyline: which other approach do you suggest then?

SimonCropp: i dont know. i dont think there is a way to solve it gracefully in MStest.

i think the only to is to pass the extra context through as a parameter to the static Verify method

TomW-Skyline: I'm wondering if walking the stack is unreliable in this case. The unit test method will always exist in the assembly, and can never be inlined.

SimonCropp: it was the approach we use in ApprovalTests. many bugs over many years. Still a few open https://github.com/approvals/ApprovalTests.Net/issues/405 https://github.com/approvals/ApprovalTests.Net/issues/449

TomW-Skyline: alright thanks, will need to solve it another way then, since we already have another base class for out tests that is not extendable.

PR #622, opened 2022-08-30
Handle empty strings properly when using strict json mode

edumserrano:
When using `VerifierSettings.UseStrictJson()` the `json` output produced is incorrect for empty strings.

Given the following model:

```csharp
public class TheTarget
{
    public string? Value { get; set; }
}
```

And the following test:

```csharp
[Fact]
public async Task Object()
{
    VerifierSettings.UseStrictJson();
    var target = new TheTarget
    {
        Value = ""
    };
    await Verify(target);
}
```

The output produced by `Verify` is:

```json
{
  "Value":
}
```

Since `UseStrictJson` is being set then the expected output should be:

```json
{
  "Value": ""
}
```


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=622) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=622) before we can accept your contribution.<br/><br/>:x: edumserrano [sign now](https://cla.dotnetfoundation.org/VerifyTests/Verify?pullRequest=622)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/VerifyTests/Verify?pullRequest=622) it.</sub>

edumserrano:
@SimonCropp I'm not sure you *like* the way I created the tests to cover the `null` or empty `string` cases. Let me know if you'd like me to change them. 

Also, I'd like to make clear that I added the test for the `null` case for completion purposes only. I found no issue with writing null values using strict mode.

edumserrano:
Thinking a bit more about this, I also noticed that if I do a test like this:

```csharp
[Fact]
public async Task EmptyStringWithPopulate()
{
    settings.AddExtraSettings(x => x.DefaultValueHandling = Argon.DefaultValueHandling.Populate);

    var target = new TheTarget
    {
        Value = ""
    };
    await Verify(target, settings);
}
```

Then `Verify` produces the expected output.

So now I'm not sure the fix I've implemented is the way this should be fixed or if this is expected behavior. Meaning even if we set `UseStrictJson` but the `DefaultValueHandling` is `DefaultValueHandling.Populate` then we don't get valid `json` output. If this is expected behavior, one thing I'd say is that since the `DefaultValueHandling` is set to `DefaultValueHandling.Ignore` other people could get into this kind of errors when using other extensions like `Verify.Quibble`.

SimonCropp:
while this PR was good, i ended up finding a few more issues that ended up conflicting with this. those changes are now deployed in 18.0.0-beta.12

edumserrano:
Awesome. Good to know the PR was good enough to describe the problem. Thank you for the quick turn around @SimonCropp.

This will let us go back to using [Verify.Quibble](https://github.com/VerifyTests/Verify.Quibble) which was how we found this problem initially. `Verify.Quibble` would fail because the JSON on the `verified` file wasn't valid.

SimonCropp:
> Verify.Quibble would fail because the JSON on the verified file wasn't valid.

that should be fixed as a side effect of my recent deployment. there is a new beta of Verify.Quibble that is compat with the new beta of Verify

Issue #658, opened 2022-09-22
Handling of paths with dots in VerifyDirectory


JanKrivanek:
## Description

` VerifyDirectory(directoryToVerify)` is not able to handle files without extensions (https://github.com/VerifyTests/Verify/blob/main/src/Verify/Verifier/InnerVerifier_Directory.cs#L69). If fixed, the extension-less files will cause issues if any of the parent folders will contain dot (https://github.com/VerifyTests/Verify/blob/main/src/Verify/Verifier/InnerVerifier_Directory.cs#L63 assumes the part after last dot to be an extension)

## Repro

**Case 1:**

```
[Fact]
public async Task VerifyDirectory_ExtensionLessFile()
{
    string dirName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName().Replace(".", string.Empty));
    Directory.CreateDirectory(dirName);
    string fileName = Path.Combine(dirName, Path.GetRandomFileName().Replace(".", string.Empty));
    await File.WriteAllTextAsync(fileName, string.Empty);

    await VerifyDirectory(dirName);
}
```

throws

```
System.ArgumentOutOfRangeException
startIndex cannot be larger than length of string. (Parameter 'startIndex')
   at System.String.ThrowSubstringArgumentOutOfRange(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex, Int32 length)
   at InnerVerifier.ToTargets(String path, Func`2 include, IEnumerable`1 enumerateFiles)+MoveNext() in /_/src/Verify/Verifier/InnerVerifier_Directory.cs:line 69
   at InnerVerifier.ToTargets(String path, Func`2 include, IEnumerable`1 enumerateFiles)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Extensions.ToList[T](IAsyncEnumerable`1 target) in /_/src/Verify/Extensions.cs:line 6
   at Extensions.ToList[T](IAsyncEnumerable`1 target) in /_/src/Verify/Extensions.cs:line 6
   at InnerVerifier.VerifyDirectory(String path, Func`2 include, String pattern, EnumerationOptions option) in /_/src/Verify/Verifier/InnerVerifier_Directory.cs:line 14
   at VerifyXunit.Verifier.<>c__DisplayClass6_0.<<Verify>b__0>d.MoveNext() in /_/src/Verify.Xunit/Verifier.cs:line 64
--- End of stack trace from previous location ---
[...]

PR #659, opened 2022-09-22
Support directories with dots and extensionless files

JanKrivanek:
## Problem

Fixes https://github.com/VerifyTests/Verify/issues/658


JanKrivanek:
@SimonCropp Appology for accidentaly including testing targeting change in .csproj, plus adding the verifications only for XUnit. Thanks for taking this to completion

SimonCropp:
@JanKrivanek no worries. thanks for the PR. this is on nuget now as a beta. i will add test few more scenarios before doing a stable deploy

Issue #654, opened 2022-09-18
Using VS remote test environment results in System.ArgumentNullException : Value cannot be null. (Parameter 'directory')


edumserrano:
## Description

Verify throws `System.ArgumentNullException : Value cannot be null. (Parameter 'directory')` when trying to run tests in Visual Studio on Windows using remote test environment targeting  `Ubuntu-20.04` via WSL2.

## How to reproduce

1) Create a new `xUnit` test project.
2) Setup a [remote test environment using WLS2](https://learn.microsoft.com/en-us/visualstudio/test/remote-testing?view=vs-2022#local-wsl2-connections).
3) Create a test that uses `Verify`.
4) Run the test.
5) Test fails with  `System.ArgumentNullException : Value cannot be null. (Parameter 'directory')` exception.

Unless this is not something you are willing to support, I would expect that instead of the exception I would get the usual `Verify` flow generating `received` files and comparing them with `verified` files.

## System

This happens when using the following NuGet package versions:

```
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.1" />
<PackageReference Include="Verify.Xunit" Version="17.10.2" />
<PackageReference Include="xunit" Version="2.4.2" />
<PackageReference Include="xunit.runner.visualstudio" Version="2.4.5">
<PackageReference Include="coverlet.collector" Version="3.1.2">
<PackageReference Include="Verify.Xunit" Version="17.10.2" />
```

in the following system:

```
.NET SDK (reflecting any global.json):
 Version:   6.0.400
 Commit:    7771abd614

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.400\
```

## More info

I also tested using the latest `Verify` pre-release version: `18.0.0-beta.18`.
It should be straightforward to reproduce this issue using the steps outlined above but if it helps, I could create a repo to reproduce this.

### testenvironments.json

This was the test environment file I used:

```json
{
  "version": "1",
  "environments": [
    {
      "name": "WSL-Ubuntu-20.04",
      "type": "wsl",
      "wslDistribution": "Ubuntu-20.04"
    }
  ]
}
```

### Verify 17.10.2 exception stack trace

```
  Message: 
    System.ArgumentNullException : Value cannot be null. (Parameter 'directory')

  Stack Trace: 
[...]

PR #700, opened 2022-11-03
Add mapping from win build paths to WSL paths

JanKrivanek:
## Problem

Running under WSL crashes due to mixing windows and subsystem mapped files
https://github.com/VerifyTests/Verify/issues/654

## Proposed solution

Added prototype mapper function mapping the build time paths to mapped WSL paths.
Utility would need to be called everywhere where build time paths are being used (so `AttributeReader` and all methods using `CallerFilePathAttribute`)

## What is missing

Adding `IoHelpers.GetMappedBuildPath` to all methods using `CallerFilePathAttribute`, only few sample usages were treated (so that debugging of `WithDirectory` test could be performed)

### DIFF
--- a/src/Verify.Xunit/Verifier.cs
+++ b/src/Verify.Xunit/Verifier.cs
@@ -37,7 +37,7 @@ public static partial class Verifier
         [CallerFilePath] string sourceFile = "") =>
         Verify(
             settings,
-            sourceFile,
+            IoHelpers.GetMappedBuildPath(sourceFile),
             _ => _.Verify(target, rawTargets));
 
     public static SettingsTask Verify(
@@ -46,7 +46,7 @@ public static partial class Verifier
         [CallerFilePath] string sourceFile = "") =>
         Verify(
             settings,
-            sourceFile,
+            IoHelpers.GetMappedBuildPath(sourceFile),
             _ => _.Verify(targets));
 
     static SettingsTask Verify(
--- a/src/Verify.Xunit/Verifier_Directory.cs
+++ b/src/Verify.Xunit/Verifier_Directory.cs
@@ -16,7 +16,7 @@ public static partial class Verifier
         object? info = null,
         FileScrubber? fileScrubber = null,
         [CallerFilePath] string sourceFile = "") =>
-        Verify(settings, sourceFile, _ => _.VerifyDirectory(path, include, pattern, options, info, fileScrubber), true);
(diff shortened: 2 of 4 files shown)


SimonCropp:
can `IoHelpers.GetMappedBuildPath` be placed deeper in the stack?

how do we test this in CI?

JanKrivanek:
Good questions - I do not have good answers though.

Deeper down the stack would require replacing of `CallerFilePathAttribute` functionality (that's detected and value injected via roslyn compiler) with custom implementation. That would probably mean harvesting the info from `StackTrace` utils - this wouldn't however work in release builds. 

CI testing - probably just a bit more abstracting and unit testing path replacing cases (and keeping identity for local run), but no easy test guaranteeing proper functionality on WSL.

SimonCropp:
> Deeper down the stack would require replacing of CallerFilePathAttribute functionality (that's detected and value injected via roslyn compiler) with custom implementation. That would probably mean harvesting the info from StackTrace utils - this wouldn't however work in release builds.

but at the time `GetMappedBuildPath` is called the `CallerFilePathAttribute` has already been injected into the `sourceFile`. that value will now be correct no matter where in the stack it is used

JanKrivanek:
> that value will now be correct no matter where in the stack it is used

Can you please elaborate more? Do you suggesting we should have a different attribute that would take care about providing properly mapped path behind the sceenes?

The `CallerFilePathAttribute` is known to the compiler and that will take care about injecting the value (https://github.com/dotnet/roslyn/blob/795612c361a8d5d9efde45f5f00da214e39a9659/src/Compilers/CSharp/Portable/Binder/Binder_Invocation.cs#L1393-L1396). Without the compiler support we cannot do any better - so the `CallerFilePathAttribute` still needs to be used. However that's not providing correct path in case of remote debugging (WSL). So in that case we need to explicitly remap somewhere in the executing code the remapping would need to be called explicitly.

But I may have very likely just misunderstood you.

Anyway - I just wanted to dump my hack to make WSL working as a sample code for possible reference in case anyone would want to play with supporting WSL remote debug scenario - not sure if worth pursuing further.

SimonCropp:
> not sure if worth pursuing further.

i think it is worth pursuing

JanKrivanek:
Superseded by https://github.com/VerifyTests/Verify/pull/729

Issue #699, opened 2022-11-03
VerifyDirectory doesn't play well with file names starting with dot


JanKrivanek:
## Problem

`VerifyDirectory` has 2 issues with file names starting with dot (e.g. `.editorconfig`):
* When file starting with dot is in the root of directory to be verified, verifier gets an exception (since file name is resolved to empty string which is disallowed):
```
System.ArgumentNullException
Value cannot be null. (Parameter 'name')
   at Guard.AgainstEmpty(String value, String argumentName) in C:\src\Verify\src\Verify\Guard.cs:line 108
   at VerifyTests.Target..ctor(String extension, StringBuilder data, String name) in C:\src\Verify\src\Verify\Target.cs:line 57
   at InnerVerifier.ToTargets(String path, Func`2 include, IEnumerable`1 enumerateFiles, Object info, FileScrubber fileScrubber)+MoveNext() in C:\src\Verify\src\Verify\Verifier\InnerVerifier_Directory.cs:line 104
   at InnerVerifier.ToTargets(String path, Func`2 include, IEnumerable`1 enumerateFiles, Object info, FileScrubber fileScrubber)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Extensions.ToList[T](IAsyncEnumerable`1 target) in C:\src\Verify\src\Verify\Extensions.cs:line 6
   at Extensions.ToList[T](IAsyncEnumerable`1 target) in C:\src\Verify\src\Verify\Extensions.cs:line 6
   at InnerVerifier.VerifyDirectory(String path, Func`2 include, String pattern, EnumerationOptions option, Object info, FileScrubber fileScrubber) in C:\src\Verify\src\Verify\Verifier\InnerVerifier_Directory.cs:line 20
   at VerifyXunit.Verifier.<>c__DisplayClass6_0.<<Verify>b__0>d.MoveNext() in C:\src\Verify\src\Verify.Xunit\Verifier.cs:line 64
```

* When the file starting with dot is in subfolder, verifier sees the verification file as an extra file and reports it as file to be deleted - this happens only on unix based OS. Repro: https://github.com/VerifyTests/Verify/pull/698

## Root cause

For the first case it's inability to work with empty string as a name. Can be solved by using null instead of empty string - code will then fallback to `NameOrTarget`.

[...]

PR #726, opened 2022-11-21
Do not create artifical name for extension-only filenames

JanKrivanek:
### Problem
Verification of files that starts with dot (e.g. `.editorconfig`) generates/uses snapshots which prepend `nofilename` to the tested file. This is a minor inconvenience - discussed here: https://github.com/VerifyTests/Verify/issues/699#issuecomment-1302881371

### Solution
filename is preserved as empty
directory structure is preserved as well
appropriate tests added

### Note
Potentially similar approach might be taken for files without extensions (currently artifical `.noextension` extension appended). This would require some changes in `EmptyFiles` as well (to support empty extension). I'm leaving it for separate work if it ever becomes concern for anyone.


JanKrivanek:
> so this is a breaking change?

That's correct.
More precisely:
* For updates from >= 18.3.1-beta.1 it's breaking
* For updates from <= 18.3.0 it's actualy enabling the upgrade *without* the breaking change that's currently imposed (as previously the 'dot filename' that was contained within some subdirectory was supported as is and in newer version the snapshot name is altered)

We are currently in the second category (trying to prevent sequence of 2 breaking changes)

SimonCropp:
do u think it makes sense for this to handle deletion of the legacy file?

if no can u add a section in the description for how people can manually clean up

JanKrivanek:
Good point.
Will look into that and update.

JanKrivanek:
@SimonCropp I update the PR to rename the snapshots from previous naming convention (simple cleanup of previous naming convention is in previous commit - if that feels as better strategy). Let me know what are your thoughts on that.
Thanks!

PR #790, opened 2023-02-18
Uniqueness list

SimonCropp:


ltrzesniewski:
@SimonCropp I was going to look into an issue I'm having but then I noticed you're modifying the same code right now. 😅

Namely: I'm getting file names like this one, using [this config](https://github.com/ltrzesniewski/InlineIL.Fody/blob/5202aa74b744b3820e4efdaef7e1a5d91b579953/src/InlineIL.Tests/Weaving/SnapshotTests.cs#L52-L56):

> AssemblyToProcess.BasicTestCases.HandleExceptionBlocks.DotNet7_0.Release.DotNet7_0.received.il

Notice that `DotNet7_0` is present twice. Looks like it's caused by this code, which duplicates logic from `PrefixUnique.SharedUniqueness`:

https://github.com/VerifyTests/Verify/blob/fef7b2f1bb1dd563d5a20b98a93315f6f4db03fb/src/Verify/Verifier/InnerVerifier.cs#L182-L186

Do you maybe intend to deduplicate prefixes following the refactor you did here?

PR #846, opened 2023-04-14
Readme about .editorconfig

moander:
I just experienced this in VS Code. The solution was what I propose in the README.


SimonCropp:
thanks

PR #855, opened 2023-04-25
Fix Mono detection in namer

mattjohnsonpint:
This restores detection of Mono when `UniqueForRuntimeAndVersion` is used, which last worked in 19.10.0, but has been broken since. #813 did not work because Mono has to be detected at runtime.  At compile time, it looks like .NET Framework.


SimonCropp:
thanks matt. is this possible to add a test for?

mattjohnsonpint:
I couldn't get the existing tests to run on my Mac at all, I think that would have to be addressed first.

Issue #884, opened 2023-05-17
Generate a hash of parameters for the file name


aaronpowell:
Recently when using Verify for a set of data-driven tests I hit a problem when providing the `UseParameters` with the parameters of the test would cause the file name and path to be excessively long and cause problems on Windows.

#### Describe the solution

To work around this, I created a method that will take the parameters and generate a hash from them instead, since that will reduce the length of the file name, but also create something that is going to be consistent across test runs.

#### Additional context

The extension can be found here: https://github.com/Azure/data-api-builder/pull/1402/files#diff-0e9ff620b043280230cfc29acae2321e450546a8b2c86b3e8e3fa289a9b36a0f

SimonCropp:
@aaronpowell i went with a slightly different approach https://github.com/VerifyTests/Verify/blob/main/docs/parameterised.md#hashing-parameters

PR #885, opened 2023-05-17
Adding parameter hashing function

aaronpowell:
This addresses the feature in #884


SimonCropp:
can u add some docs

aaronpowell:
> can u add some docs

Docs added with sample snippet.

SimonCropp:
@aaronpowell did u end up using this feature?

aaronpowell:
I'd assume so, but I don't know if it's still in use

SimonCropp:
@aaronpowell can u check? context https://github.com/VerifyTests/Verify/pull/1449

if u still have a use case, i would like to dive deaper into it to work out if there is a better way to support it. or, if not, i might pull out the hashing to a diff nuget so it can move forward with the supported runtime of System.IO.Hashing

PR #911, opened 2023-06-02
Document workaround for serializing F# lists of DUs

mlaily:
Hello,

This is unfortunate, but it looks like the order json converters are added matters, and the F# discriminated unions converter is added last, so trying to serialize a list of DU types fails with an error similar to this:
```
Argon.JsonWriterException : Unsupported type: REDACTED.Business+CommandWithProgress. Use the JsonSerializer class to get the object's JSON representation. Path ''.
```

Manually adding the `DiscriminatedUnionConverter` in the extra converters just before the other F# converters fixes the issue.

### DIFF
--- a/docs/mdsource/fsharp.source.md
+++ b/docs/mdsource/fsharp.source.md
@@ -8,6 +8,13 @@ To serialize F# types properly, add the converters:
 VerifierSettings.AddExtraSettings(fun settings -> settings.AddFSharpConverters())
 ```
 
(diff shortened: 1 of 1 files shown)


mlaily:
Hey sorry, I think I was too quick: adding the DU converter first fixed the exception that I was seeing in my test, but the result is actually still incorrect!

(I'm getting a json representation of nested nodes from my fsharp linked list, instead of a normal json array)

I'm going to close this PR and open another one in Argon.

Sorry again for the inconvenience.

PR #930, opened 2023-07-04
XML docs: param → paramref where appropriate

ForNeVeR:
Otherwise, the `path` gets lost from IDEs like Rider :(

Before:
image

After:
image


SimonCropp:
thanks

SimonCropp:
is Jetbrains using Verify?

ForNeVeR:
I am not aware of such usage.

I personally use Verify, though 😅

Issue #935, opened 2023-07-13
VerifyXml method does not work correctly with empty tags


callmekoo:
Method VerifyXml expects opening and closing tags instead of empty tag.

Test to reproduce:
```
[Fact]
public Task EmptyTag() => VerifyXml("<body><empty /><node>text</node></body>");
```
This test fails with error, received file contains this xml:
```
<body>
  <empty></empty>
  <node>text</node>
</body>
```
Obviously, this is not correct xml, the empty tag should be a single tag.

The bug has been reproduced since version 19.8.2, in older versions everything works correctly.

SimonCropp:
do you want to have a go at fixing this?

callmekoo:
I'll try it this weekend, but I'm not quite sure of my skills.

PR #938, opened 2023-07-15
Bugfix for VerifyXml

callmekoo:
Method VerifyXml doesn't work correctly with empty xml tags as described in #935. Here is a fix for this problem.


SimonCropp:
@callmekoo can u accept the cla

callmekoo:
@dotnet-policy-service agree

PR #972, opened 2023-08-24
Add `Verify(Func<Task<T>>)`

dahlbyk:
A few times now we've wanted to make a helper like this, returning `SettingsTask` so we can `UseParameters()` or whatever:

```csharp
static SettingsTask VerifyFoo(Foo foo, BarRepo bars) =>
    Verify(new
    {
        foo,
        bar = await bars.GetBars(foo), // requires async!
    });
```

But the `await` requires `async Task<SettingsTask>` which then requires `await await VerifyFoo(...)`. Not pretty.

Accepting `Func<Task<T>>` makes building async targets much cleaner:

```csharp
static SettingsTask VerifyFoo(Foo foo, BarRepo bars) =>
    Verify(async () => new
    {
        foo,
        bar = await bars.GetBars(foo),
    });
```


SimonCropp:
can you add some docs for this scenario

dahlbyk:
> can you add some docs for this scenario

Added an **Async** section before **JsonVerify**.

dahlbyk:
I guess I could also explicitly document the pattern one wants to wrap a `Verify()` call in a helper method like we did. Thoughts?

SimonCropp:
> I guess I could also explicitly document the pattern one wants to wrap a Verify() call in a helper method like we did. 

i dont think it is necessary. 

i will merge and deploy this now.

sorry for the delay

dahlbyk:
> sorry for the delay

No worries at all. Thanks for the library!

Issue #635, opened 2022-09-02
Add options for BOM and EOF newline


mattjohnsonpint:
When Verify generates its "received" files, it always creates a BOM at the top of the file, and it never adds a newline at the end.  These both show with red glyphs on GitHub.

One can remove the BOM and add a newline manually in the "verified" file, and those are ignored by Verify when comparing, but diff tools will still show them if there are other changes in the file.  Also it's a pain if you just want to accept the whole file.

Since people have different preferences in this regard, I propose addings some settings to control them when Verify generates the received file.

I'd also suggest the defaults be UTF8 *without* BOM, and always a newline at the end of the file.  But that's just my preference.

There could also be an option for line endings, but generally I think this is handled fine by using `Environment.Newline` and/or by git itself.

SimonCropp:
the BOM is because, when it is omitted, many diff tools incorrectly guess the encoding. if we add an option to omit it, other people will hit the encoding issue

as for the trailing newline, i dont quite understand what the friction is?

nightroman:
> There could also be an option for line endings, but generally I think this is handled fine by using Environment.Newline and/or by git itself.

I'm second to this. We hit an issue - an original and committed `verified` is produced on Windows. But then a test fails on Mac due to the different `received`. The difference is only line ends.

SimonCropp:
@nightroman 

> I'm second to some option "ignore line endings". We hit an issue - an original and committed verified and produced received differ only by line endings. And the check failure is unwanted in this case.

this is a bug. verify should ignore line ending differences. can u push up a repro to a GH repo so i can take a look

note that the choice of lf as a line ending was done for perf. it means that on reading a verified file, it can be read as a single string instead of split into lines and re-joined. 

i have found that [setting the git line endings](https://github.com/VerifyTests/Verify#source-control-received-and-verified-files) mitigates any line ending mis match issues.

nightroman:
> i have found that [setting the git line endings](https://github.com/VerifyTests/Verify#source-control-received-and-verified-files) mitigates any line ending mis match issues.

As far as this is documented, it is rather not a bug.
We ourselves ended up setting this in .gitattributes.

So after all I'm not sure about my request about adding "ignore line endings" options.

But without hurting performance too much, only when files are different, Verify may check for line endings and fail with an error about not supported line endings if found.

SimonCropp:
> without hurting performance too much, only when files are different, Verify may check for line endings and fail with an error about not supported line endings if found

[...]

PR #1074, opened 2023-11-26
Add support for disabling UTF-8 BOM

joelverhagen:
Based on https://github.com/VerifyTests/Verify/issues/635#issuecomment-1546619357, there appears to be an opportunity for disabling UTF-8 BOM on snapshot files. I made a PR for this. I added a more general `settings.UseEncoding(...)` method also so people can customize the encoding further or revert back to UTF-8 with BOM.

I added some tests and marked the snapshot files as binary in `.gitattributes` to avoid encoding problems on these particular test files.


src/Verify/VerifySettings.cs
```diff
@@ -88,6 +89,14 @@ public void UseTextForParameters(string parametersText)
         this.parametersText = parametersText;
     }
 
+    public void UseUtf8NoBom() =>
```
SimonCropp, line 92: I dont see people wanting an instance API for this? wouldnt they only want the static API?

src/Verify/VerifierSettings.cs
```diff
@@ -30,4 +30,12 @@ public static void AutoVerify(bool includeBuildServer = true)
     }
 
     internal static bool autoVerify;
+
+    public static void UseUtf8NoBom() =>
```
SimonCropp, line 34: is there a test for this?

src/Verify/VerifySettings.cs
joelverhagen: Not sure. I want the static API indeed but I didn't know the criteria used for one or the other. Let me know and I can remove it.

src/Verify/VerifierSettings.cs
joelverhagen: I was concerned about calling the static method in the test since it might impact other tests. Is there a good way for me to isolate this execution? Perhaps it's a solved problem in your test suite.

src/Verify/VerifySettings.cs
SimonCropp: the instance API would only be if someone wants to mix BOM and non-bom in the same project. i dont think that would eve happen. so lets go with only the static API

src/Verify/VerifierSettings.cs
SimonCropp: you will need to add a new test project to achieve the isolation

src/Verify/VerifySettings.cs
joelverhagen: sounds good!

src/Verify/VerifierSettings.cs
joelverhagen: Done. Worked locally. Watching the CI.

src/Verify/VerifySettings.cs
joelverhagen: done.

SimonCropp:
while the CI does its thing. what project are you using Verify on?

joelverhagen:
> while the CI does its thing. what project are you using Verify on?

NuGet Insights. My notes on the possibility are here if you're interested https://github.com/NuGet/Insights/issues/100.

Overall I like the workflow of Verify more than the snapshot testing I "discovered" in a hacky way on my own. I think snapshot testing is a great tool for ETL-like flows implemented by NuGet.org.

joelverhagen:
Awesome, thanks for that fast merge! Great project!

SimonCropp:
this is now deployed. thanks for the contribution

Issue #1081, opened 2023-11-29
Allow file extension to be part of the unique snapshot check


joelverhagen:
#### Describe the bug

Verify acts in a strange way when there are two `Verify` calls with the same snapshot prefix but a different file extension. I was expecting file extension to be part of the "identity" of the snapshot file (as the file system would allow) but this appears to not be the case. The snapshot gets renamed based on whichever `Verify(..., extension: "csv or json")` goes last.

Apologies if this should be a feature request instead of a bug report. It _feels_ buggy to me but that's perhaps just due to a bad expectation on my part.

I ran into this because I have a JSON file called `..._PackageReadmes.verified.json` and a CSV called `..._PackageReadmes.verified.csv` in the same test file. It ran into this problem.

#### Minimal Repro

```csharp
namespace VerifyExt;

[UsesVerify]
public class Tests
{
    [Fact]
    public async Task MyTest()
    {
        VerifierSettings.AutoVerify();
        VerifierSettings.DisableRequireUniquePrefix();

        await Verify("foo", extension: "csv").UseTextForParameters("MyFoo");

        // 'Tests.MyTest_MyFoo.verified.csv' is on disk right now

        await Verify("bar", extension: "txt").UseTextForParameters("MyFoo");

        // Now it is 'Tests.MyTest_MyFoo.verified.txt'. The CSV is gone!
    }
}
```

#### Submit a PR that fixes the bug

TODO. I will work on this now. I think I will try adding an option `IncludeExtensionInUniquePrefix` to take the stance that this is by design instead of a bug. Let me know if that's the wrong direction.

SimonCropp:
can you instead use this approach>

    [Fact]
    public async Task MyTest()
    {
        await Verify(new List<Target>
        {
            new("csv", "foo"),
            new("txt", "bar")
        });
    }

joelverhagen:
I tried this in my test suite and face the same problem. It appears to be a race condition based on what happens before each Verify call. I had some async work reading from Azure Storage inside async helper methods that at the end call Verify.

There appears to be some sync setup in the Verify task. If any of this is interleaved then the behavior is inconsistent. I'll confirm this theory by switching the order of the Verify calls and the subsequent awaits.

joelverhagen:
[...]

PR #1082, opened 2023-11-29
Add UniqueForFileExtension to allow snapshots with different extension to coexist 

joelverhagen:
Fix https://github.com/VerifyTests/Verify/issues/1081


SimonCropp:
how would this work given a single call to Verify can produce multiple files with different file extensions?

joelverhagen:
Is there another approach I should take with this PR? I'd like to use file extension as a differentiator in a snapshot file instead of having the verify process delete these different files.

SimonCropp:
what if i changed the behavior to "multiple test cannot produce the same verified file"  ?

joelverhagen:
> what if i changed the behavior to "multiple test cannot produce the same verified file" ?

This sounds like a "block" rather than "allow" behavior. For that idea, what would happen if Verify call 1 produces `foo.verified.txt` and Verify call 2 produces `foo.verified.csv`? Verify call 2 would fail? I think I'm a little unclear on what that means.

I'd be happy to give it a try if you have a POC or can guide me to implementing the behavior you're thinking of!

SimonCropp:
> For that idea, what would happen if Verify call 1 produces foo.verified.txt and Verify call 2 produces foo.verified.csv? Verify call 2 would fail?

no it would not fail

i will put together the change and let u know tomorrow

aradalvand:
+1 just hit this exact use case.

aradalvand:
+1 just hit this exact use case.

SimonCropp:
closing this one. I suggest instead you submit a PR that meets your requirements with a single Verify call

PR #1098, opened 2023-12-22
also scrub guid, date, and time string keys

### LINKED ISSUE #1096: Guids as string keys not being scrubbed.
Note: New issues raised, where it is clear the submitter has not read the issue template, are likely to be closed with "please read the issue template". Please don't take offense at this. It is simply a time management decision. If someone raises an issue, and can't be bothered to spend the time to read the issue template, then the project maintainers should not be expected to spend the time to read the submitted issue. Often too much time is spent going back and forth in issue comments asking for information that is outlined in the issue template.

#### Describe the bug

I have a custom dictionary class and a custom ID struct that is 90% of the time a struct. Everything works great and Guids are in general handled properly. 

But I have a guid in dictionaries where it does not work. I found a 2 year old bug about that:
https://github.com/VerifyTests/Verify/issues/74

I tried to dig into the code, but I am not sure how you have actually solved it. For my understanding it should be here: https://github.com/VerifyTests/Verify/blob/main/src/Verify/Serialization/VerifyJsonWriter.cs#L63. I can also not find a custom converter for dictionaries. Perhaps this is a regression...

Here is my screenshot from the merge tool.
image

#### Minimal Repro

Happy to provide this if needed.

Submit a [Pull Request (PR)](https://help.github.com/articles/about-pull-requests/) that fixes the bug. Include in this PR a test that verifies the fix. If you were not able to fix the bug, a PR that illustrates your partial progress will suffice.

SimonCropp:
fixes #1096


SebastianStehle:
I think that it still does not work 100% as expected. I think many of the scrubbing works on the JSON level, so if you have a custom type that works like a date (e.g. Node.Instant), it works as expected. In my case it is a `DomainId` that can be a Guid. 

But in the contract resolver it works on the CLR types.

SimonCropp:
@SebastianStehle if u can share a failing test as a PR i can check if it is feasible to work OOTB

SebastianStehle:
Sure, I can do that after Christmas. I think the use case would he a type that is serialized to a string and either represents a datetime or guid. Especially value types that are convertible to strings seem to be handled by argon already

SimonCropp:
does ScrubInlineGuids work for you ?https://github.com/VerifyTests/Verify/blob/main/docs/guids.md#inline-guids

Issue #1034, opened 2023-10-23
Ability to specify sub directory for each call to .Verify()


MJB222398:
#### Is the feature request related to a problem

I want to verify a zip file containing .xml files, some .xml files are at the root of the zip and others are in sub directories in the zip file. My first thought was to use VerifyZip() method. But this I think just does a text comparison on each file and I would really like to have the functionality that VerifyXml() gives - especially the pretty print. So I played around with iterating over the zip entries and using VerifyXml() on those. My problem with that though is I cannot recreate the directory structure of the zip file if I am using a custom directory for all my Verify artifacts. The VerifyZip method did perfectly keep the structure of the zip file so its clearly possible for Verify to do this. So I have the following set globally:

```
Verifier.DerivePathInfo((sourceFile, projectDirectory, type, method) =>
    new(
        directory: Path.Combine(projectDirectory, Snapshots.SnapshotsDirectoryName, Snapshots.DomDirectoryName, type.Name, method.Name),
        typeName: type.Name,
        methodName: method.Name));
```

I can't then seem to create sub-directories inside of here for my zip file entries. Tried using .UseDirectory() but that overwrites my directory structure defined above:

```
foreach (ZipEntry file in zipArchive)
{
    using var fileStream = zipArchive.GetInputStream(file);

    await Verifier.VerifyXml(fileStream)
        .UseDirectory(Path.GetDirectoryName(file.Name))
        .UseFileName($"{Browser}_{Path.GetFileName(file.Name)}");
}
```

Am I using Verify wrong, or maybe missing something that I could use instead? Or perhaps it is possible to use VerifyZip and use VerifyXml on each xml file within that?

#### Describe the solution

If I cannot use VerifyZip together with VerifyXml then perhaps the addition of a sub-directory method would be the solution, where this maintains the existing directory settings and just creates a sub directory within that for the file in question. Something like:

```
foreach (ZipEntry file in zipArchive)
{
    using var fileStream = zipArchive.GetInputStream(file);

    await Verifier.VerifyXml(fileStream)
        .UseSubDirectory(Path.GetDirectoryName(file.Name))
        .UseFileName($"{Browser}_{Path.GetFileName(file.Name)}");
}
```

#### Additional context

For transparency the 'zip' is actually an Excel ,xlsx file. So ideally I would compare the file as such but doing so requires licences that my employer will not purchase.

SimonCropp:
> For transparency the 'zip' is actually an Excel ,xlsx file

[...]

PR #1126, opened 2024-01-08
Allow sub directory to be configured

MJB222398:
To implement functionality in #1034


MJB222398:
> @dotnet-policy-service agree

MJB222398:
@dotnet-policy-service agree

PR #1141, opened 2024-02-07
Allow properties to be included even when they are the default value

### LINKED ISSUE #1140: Add a way to include properties which are equal to their default value

The problem is that there is currently no good way to include specific types/properties in the Verify output if they are equal to their default value. It is possible to set `DefaultValueHandling = DefaultValueHandling.Include` for the serializer settings at a global level, but generally this is unwanted as most properties should indeed be excluded if they are the default.

However it's quite useful for specific properties or types to be included regardless to keep the verify output uniform and clean.
For example, my main use case is for numbers that might fluctuate between zero and numbers close to zero. The verify properties popping in and out is problematic. In addition, we employ a value converter to round the values which allows small insignificant changes in floating point results to pass the verify tests. However, the value converter isn't even invoked when the value is zero leading to failing tests.

#### Describe the solution

My proposed solution is to add a new value to the `ScrubOrIgnore` enum called `AlwaysInclude` or similar which will force the property to be included by the serializer even if it is the default value. See #1141 

One alternative is to make the `CustomContractResolver` public so it could be subclassed. This is a simpler change but leaves the burden of functionality on the library user, and also creates a new public-facing interface. I'd be satisfied with this change if the suggested solution isn't acceptable.

JustASquid:
Resolves #1140 

Haven't added documentation yet, awaiting approval on the API.
Right now this only allows granularity on the type level; if relevant we could add the other levels of granularity similar to ignore/scrubbing - want to get the basic API approved first.


JustASquid:
@dotnet-policy-service agree company="ACT Building Systems"

SimonCropp:
@JustASquid i am happy with the API. can you add some documentation for the feature

SimonCropp:
oh. also the test is duplicated into the StrictJsonTests. so u need to accept the result from that project as well

JustASquid:
Thanks @SimonCropp I was going to ask why the tests were failing - that makes sense.

I will add some documentation ASAP.

PR #1151, opened 2024-02-24
Use await using

mokarchi:
Similar as using (...) uses IDisposable to clean up resources, await using (...) uses [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable?view=dotnet-plat-ext-3.0). This allows to perform also time-consuming tasks (e.g involving I/O) on cleanup without blocking.


mokarchi:
The problem is that in .NET Standard 2.0, Stream doesn't implement IAsyncDisposable.

PR #1168, opened 2024-03-24
Add StringSyntaxAttribute

SimonCropp:
So we get better intellisense 
Untitled


SeanFeldman:
And can be [added](https://steven-giesel.com/blogPost/d2b2fc18-ca4c-4879-b87f-a1d36f435805) to earlier versions of .NET

SimonCropp:
@SeanFeldman yep. polyfill means it works in the following
image


https://github.com/SimonCropp/Polyfill?tab=readme-ov-file#stringsyntaxattribute

PR #1183, opened 2024-04-10
Add *.received.* to MSTest's results attachments

### LINKED ISSUE #1182: [Feature Request] Attach .received. files to test context for MSTest

As stated in the getting started wizard, debugging test failures usually requires updating the build pipeline YAML to upload *.received.* files as logs / artifacts.
 
#### Describe the solution

MSTest has support for attaching "result files" (see https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.testcontext.addresultfile?view=visualstudiosdk-2022). It would be nice if failing tests attached the corresponding `*.received.*` file to make it easier to view / diff failures without needing to modify the build pipeline YAML.

The result is that the received file is available to view / download like this: https://learn.microsoft.com/en-us/azure/devops/pipelines/test/review-continuous-test-results-after-build?view=azure-devops#troubleshooting-data-for-test-failure

The alternative is likely do-nothing. Motivated people can modify their build pipelines to collect and upload received files on failure.

-----

I'm happy to do the work here, but I'm not sure where this code best belongs. I think the idea would be to catch exceptions in the TestBase::Verify call, attach the file, and then re-throw, but there may be a better option.

If you have any questions / concerns, please let me know. Thanks!

MattKotsenas:
Fixes #1182

This change adds any *.received.* files from failing diffs to MSTest's results attachments. Attaching the received file makes it easy to download from a CI server's test results viewer without needing to modify the CI pipeline or YAML files to collect received files and upload them as build artifacts.

To accomplish that I:

1. Updated the `VerifyException` to track the `FilePair`s for mismatches
2. Wrap the inner verify to catch the exception and add the FilePair to the `TestContext`

In order to test the behavior, the logic is as follows:

1. Create a test where the *.received.* will not match the *.verified.* (Foo and Bar in this case)
2. Disable diffing so that unit test runs don't pop a diff tool
3. Create a custom `[TestMethod]` attribute, as that's the only way I know of to access the attachments
4. As part of `ReceivedFileRequiredTestMethodAttribute`'s execute, verify each test has a *.received.* file, and if not set the test result to failed


MattKotsenas:
Opening this more to start a conversation to decide if this feature has merit. If yes, then I'm open to implementation alternatives. The current PR uses a decent amount of LINQ, but since it's in failure cases I opted for clarity to start.

If you have any questions / concerns, please let me know. Thanks!

MattKotsenas:
One alternative to possibly consider is doing the attachment in a custom test attribute instead of in the "core" verify call.

However I've found that custom test attributes:

- tend to have low discoverability
- don't compose well with other custom attributes

SimonCropp:
i am currently thinking of taking a diff approach https://github.com/VerifyTests/Verify/pull/1186/files

PR #1193, opened 2024-04-13
Use attribute instead of base class for MSTest

### LINKED ISSUE #1184: [Feature request] Support MSTest without requiring VerifyBase base class

Currently, MSTest support uses a `VerifyBase` base class that all tests must inherit from. This requirement can cause issues with other test tools that also use a base class for extensibility since tests can only have one base class.

From what I can tell, the only reason to need a base class specifically is to declare the `TestContext` property, which MSTest injects.

#### Describe the solution

The feature request is to support an alternative way of adding Verify support to MSTest classes that does not consume the base class.

One solution that seems workable to me would be to introduce an attribute such as `[UsesVerify]` paired with a source generator that adds the TestContext property to the class. The `Verify()` calls could then be supplied via a static using like they are for xunit.

Unsure of alternative solutions, but open to other ideas.

#### Additional context

I'm happy to do the work myself if you'd like, but I wanted to ask before starting in case a feature like this was already considered and rejected.

If you have any questions / concerns, please let me know. Thanks!

MattKotsenas:
Fixes #1184 

## Overview

Introduce an attribute `[UsesVerify]` for MSTest so that tests can set up Verify without needing to use the `VerifyBase` class. Relaxing the base class requirement makes it easier to use Verify in more cases, where other test scenarios may also require using a base class.

The attribute triggers a source generator that adds the required `TestContext` property. The setter then stashes the context on an async local for use by the Verifier.

## Highlights

- Created a static `Verifier` class like the other test frameworks do and added it to the implict usings
- Removed logic from `VerifyBase` and forward calls to `Verifier`
    - This means we can only rely on the information in `TestContext` to get the `Type` and `MethodInfo` used for deriving path info
    - For better performance the Verifier first tries reflection to get the assembly of the test class, and then falls back to assembly scanning
- Created source generator that adds the TestContext property that sets the AsyncLocal on Verifier
    - Unit tests intentionally do not use the Verify.SourceGenerators to avoid a circular dependency between repos
- Moved all existing unit tests from `VerifyBase` to `[UsesVerify]`
    - Added a simple set of tests to verify the base class doesn't regress
- Added a dependency on ViHo.PackAsAnalyzer
    - This is the same logic the runtime repo uses to package analyzers / source generators into the main package

## Open Issues

- [x] This PR is big and (in my opinion) difficult to review; do you want me to split it up into preparatory refactoring changes?
    - [x] Calculate name and path info from TestContext and stop relying on `VerifyBase` inheritance
    - [x] Create the Verifier class and forward calls from VerifyBase
    - [x] Create the actual source generator & update tests
- [x] Agree on reflection / strategy for finding test Type and MethodInfo
- [x] Agree on if we want to allow hooking the TestContext setter
- [x] Agree on using `ForAttributeWithMetadataName`, which limits users to .NET 7 SDK+ (but can still use any supported TFM)
    - [x] Where should this be documented?
- [x] Are there other scenarios that should have `VerifyBase` tests?


MattKotsenas:
@SimonCropp , let me know if there's a better way to collaborate here. There's a few design questions to discuss, so I thought I'd start a draft PR early, and we can iterate here.

If you have any questions, please let me know.

src/Verify.MSTest.Tests/Tests.cs
```diff
@@ -1,14 +1,29 @@
-﻿// ReSharper disable UnusedParameter.Local
+﻿// TODO: Add a test that uses the base class to prevent regressions until it's deleted
+// TODO: Add a test for when the [TestClass] already has a TestContext property
+
+#region WillBeSourceGenerated
+partial class Tests
```
MattKotsenas, line 5: Reviewer note: In order to get feedback sooner, I manually generated what the source generator will.

src/Verify.MSTest/Verifier.cs
```diff
@@ -0,0 +1,138 @@
...
+        return new(
+            sourceFile,
+            settings,
+            type.NameWithParent(),
+            method.Name,
+            method.ParameterNames(),
+            pathInfo);
+    }
+
+    static Type FindType(ReadOnlySpan<char> typeName)
```
MattKotsenas, line 67: @SimonCropp , if we switch MSTest to the other DerivePathInfo form that takes strings for the type and method name instead of `Type` and `MethodInfo`, we can avoid all this assembly scanning. However, that's presumably a breaking change for consumers.

Should we keep the signature the same for compatibility, or change for performance? If we need to keep it the same, I'll add some caching here to prevent the repeated reflection lookups.

src/Verify.MSTest.Tests/Tests.cs
```diff
@@ -1,14 +1,29 @@
...
+{
+    [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "Injected by the test framework per test.")]
+    public TestContext TestContext
```
MattKotsenas, line 8: @SimonCropp the downside of the MSTest extensibility model is that we need to create this property in the _user's_ class, so there could be compilation errors if they already have the property defined. I think this is ok because `VerifyBase` also defined this property, so people moving from the base class already couldn't define this property themselves.

However, if users also wanted to hook get or set like we are, they couldn't with my simple code. I _could_ create a private setter method to update the async local, and if there's a conflict document that users should call our helper, however I don't like that because:

1. I don't think it's a very common scenario
2. That would then require a private property to store the TestContext, and add yet another field to the user's test class, which feels dirtier.

For reference that hook-able generated code would look something like this

```csharp
partial class Tests
{
    private TestContext? __verifyTestContext;

    public TestContext TestContext
    {
        get => __verifyTestContext;
        set
        {
            __verifyTestContext = value;
            OnSetTestContext(value);
        }

        private static void OnSetTestContext(TestContext context)
        {
            CurrentTestContext.Value = context;
        }
}
```

SimonCropp: >  I think this is ok because VerifyBase also defined this property, so people moving from the base class already couldn't define this property themselves.

but can they use the generated one? ie in the base class case, if they want to leverage TestContext they could use the one from the base.

SimonCropp: what if we name the generated property something that wont conflict eg `public TestContext VerifyTestContext`

src/Verify.MSTest/Verifier.cs
SimonCropp: happy to make this a breaking change

lets go with strings. if someone has legit requirements for type and method, we can revisit at that point

MattKotsenas:
I haven't forgotten about this. I got pulled into since stuff at work and have to put this old hold for another week or two.

src/Verify.MSTest/Verifier.cs
MattKotsenas: After several hours I wasn't able to find a combination of information directly available on the `TestContext` that matched the behavior before and passed the existing tests.

As a result, I left in the assembly scanning as a fallback, but first try to grab the type from the `TestContext` internals using reflection. I dislike that solution but wasn't able to find anything I liked better. Open to suggestions here.

src/Verify.MSTest.Tests/Tests.cs
MattKotsenas: I can't name the generated property something different, as MSTest is doing property injection by name.

To start I'm only generating the property if one doesn't already exist and relying on the exception message in `Verifier` to help correct any issues a user might have.

Also note that this would be limited to users that need to hook the _set_. If they only want to get the TestContext they can call our generated property.

If you disagree with that plan let me know and I can use the above technique to give users a hook.

SimonCropp:
i tried to build and got
```
UsesVerify.g.cs(16,19): Error CS0542 : 'VerifyBase': member names cannot be the same as their enclosing type
16>UsesVerify.g.cs(18,28): Error CS0548 : 'VerifyBase.VerifyBase.TestContext': property or indexer must have at least one accessor
16>CSC: Error CS8785 : Generator 'UsesVerifyGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'ArgumentOutOfRangeException' with message 'repeatCount ('-4') must be a non-negative value. (Parameter 'repeatCount')
1
```

note i am using Rider. Any ideas?

SimonCropp:
> This PR is big and (in my opinion) difficult to review; do you want me to split it up into preparatory refactoring changes?

no need to split. it is a big PR. but many of the changes are not significant

> Agree on reflection / strategy for finding test Type and MethodInfo

happy with reflection. can UnsafeAccessor with a `#if` be used? 

> Agree on if we want to allow hooking the TestContext setter

can you expand on this

> Agree on using ForAttributeWithMetadataName, which limits users to .NET 7 SDK+ (but can still use any supported TFM). Where should this be documented?

I will include .NET 7 SDK+ in the main doco. so u dont need to add it as part of this PR

> Are there other scenarios that should have VerifyBase tests?

IMO the current tests are sufficient

SimonCropp:
ignore my build error. a restart of rider fixed it

MattKotsenas:
> ignore my build error. a restart of rider fixed it

Yeah, this is a downside of using a generator (or analyzer) as a ProjectReference; it's loaded into the build server and thus old versions can be cached, see https://github.com/dotnet/roslyn/issues/48083

These problems _should_ only happen the first time the generator is added to a project and won't impact users since they won't be changing/building the generator as part of their build.

MattKotsenas:
Regarding hooking the TestContext: I think this is a case where I was overcomplicating things.

_In theory_ a user could have a test setup where they also want to do some custom action when the TestContext was set. Let's say they had code like this:

```csharp
private TestContext _localField;

public TestContext TestContext
{
    get { return _localField; }
    set
    {
        Console.WriteLine(value);
        _localField = value;
    }
}
```

because the generator needs to write this property to set `Verifier.CurrentTestContext.Value`, there's no easy way for a user to hook the set.

We _could_ support this scenario one of two ways:

1. Updating our source generation to look more like this

```csharp
public TestContext TestContext
{ 
    get;
    set
    {
        CurrentTestContext.Value = value;
        VerifyOnSetTestContext(value);
    }

    private virtual void VerifyOnSetTestContext(TestContext context)
    {
    }
}
```

and the user could implement `public override void VerifyOnSetTestContext(TestContext context) => Console.WriteLine(context);`

2. We could document `Verifier.CurrentTestContext` and tell the user to instead of using `[UsesVerify]` to set the async local themselves

No existing code using VerifyBase could be in this situation since the property was defined on the base.

So in summary I'm not sure this is a use case worth supporting without someone asking. My recommendation would be to keep the source generator as-is. If you think this is a scenario worth supporting (or someone asks for it) I'd prefer option 1 over option 2 since it provides a cleaner contract with the user.

Let me know if I'm still not making sense :)

MattKotsenas:
OK, I think I'm good with the changes here. Let me know if there's anything else from your side.

Thanks for the review and all the help!

SimonCropp:
>So in summary I'm not sure this is a use case worth supporting without someone asking. My recommendation would be to keep the source generator as-is. If you think this is a scenario worth supporting (or someone asks for it) I'd prefer option 1 over option 2 since it provides a cleaner contract with the user.

ok. happy with leaving as is for now.

head up: I am going to push some minor cleanups

src/Verify.MSTest.SourceGenerator/ClassToGenerate.cs
```diff
@@ -0,0 +1,13 @@
+readonly record struct ClassToGenerate
+{
+    public string? Namespace { get; }
+    public string ClassName { get; }
+    public EquatableArray<ParentClass> ParentClasses { get; }
```
SimonCropp, line 5: cant we rely on the gethashcode and equality generated methods that we get via record struct? then we dont need EquatableArray

src/Verify.MSTest.SourceGenerator.Tests/Tests.HasAttributeInNamespace.verified.txt
```diff
@@ -0,0 +1,25 @@
...
+  Item2:
+//-----------------------------------------------------
+// This code was generated by a tool.
+//
+// Changes to this file may cause incorrect behavior
+// and will be lost when the code is regenerated.
+// <auto-generated />
+//-----------------------------------------------------
+
+namespace Foo
```
SimonCropp, line 12: is it possible to use file scoped namespaces?

src/Verify.MSTest.SourceGenerator.Tests/Tests.cs
```diff
@@ -0,0 +1,137 @@
...
+    readonly UsesVerifyTestDriver testDriver = new();
+
+    async Task VerifyGenerator(GeneratorDriverResults results)
+    {
+        output.WriteLine($"First run of generators took: {results.FirstRun.TimingInfo.ElapsedTime}");
+        output.WriteLine($"Cached re-run of generators took: {results.CachedRun.TimingInfo.ElapsedTime}");
+
+        await Verify(results.FirstRun.RunResult.SelectGeneratedSources());
+
+        // Ensure cachability
```
SimonCropp, line 16: for my own education. what make it cacheable? is it equality? or hashcodes? or both?

src/Verify.MSTest.SourceGenerator/Parser.cs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    static bool HasTestContextPropertyDefinedInBase(INamedTypeSymbol symbol)
+    {
+        static bool HasTestContextProperty(INamedTypeSymbol symbol) =>
+            symbol
+            .GetMembers()
+            .OfType<IPropertySymbol>()
+            .Any(_ => _.Name == "TestContext" &&
+                      _.Type.Name == "TestContext");
```
SimonCropp, line 58: should we check namespace?

src/Verify.MSTest.SourceGenerator/TrackingNames.cs
```diff
@@ -0,0 +1,17 @@
+namespace VerifyMSTest.SourceGenerator;
+
+public static class TrackingNames
+{
+    public static string InitialTransform => nameof(InitialTransform);
+    public static string RemoveNulls => nameof(RemoveNulls);
+    public static string Collect => nameof(Collect);
+
+    public static IReadOnlyCollection<string> GetTrackingNames() =>
+        typeof(TrackingNames)
```
SimonCropp, line 10: wouldnt this be simpler as a hard coded array of 3 strings?

src/Verify.MSTest/Verifier.cs
```diff
@@ -0,0 +1,215 @@
...
+    /// As an optimization, try to retrieve the <see cref="ITestMethod"/> stored on the
+    /// <see cref="TestContext"/>, and use that to retrieve the <see cref="Type"/>.
+    ///
+    /// If reflection fails, return <c>false</c>.
+    /// </summary>
+    /// <param name="typeName">The fully qualified name of the class of the currently running test.</param>
+    /// <param name="testContext">The <see cref="TestContext"/> of the current test.</param>
+    /// <param name="type">The <see cref="Type"/> of the currently running test.</param>
+    /// <returns><c>true</c> if the reflection succeeded; <c>false</c> otherwise.</returns>
+    static bool TryGetTypeFromTestContext(string typeName, TestContext testContext, [NotNullWhen(true)] out Type? type)
```
SimonCropp, line 93: why so defensive in this method. IMO it should assume the correct shape and throw if that shape is not correct

src/Verify.MSTest/Verifier.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        try
+        {
+            // We can't use UnsafeAccessor in this context because _testMethod is of type
+            // Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.ObjectModel.TestMethod, which isn't part of the MSTest.TestFramework
+            // package. If / when an API like https://github.com/dotnet/runtime/issues/90081 lands this generic reflection can be replaced.
+            var testMethod = testContext
+                .GetType()
+                .GetField("_testMethod", BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
```
SimonCropp, line 102: can we cache this field

src/Verify.MSTest/Verifier.cs
```diff
@@ -0,0 +1,215 @@
...
+                ?.GetValue(testContext);
+            var assemblyPath = testMethod
+                ?.GetType()
+                ?.GetProperty("AssemblyName", BindingFlags.Instance | BindingFlags.Public)
```
SimonCropp, line 106: can we cache this property

src/Verify.MSTest.SourceGenerator/TrackingNames.cs
MattKotsenas: Yes. I was trying to avoid the case where someone added a new name was added but forgot to update the array. However, that's probably not worth it. Updated to a simple array.

src/Verify.MSTest.SourceGenerator/Emitter.cs
```diff
@@ -0,0 +1,79 @@
...
+    static readonly string GeneratedCodeAttribute =
+        $"[global::System.CodeDom.Compiler.GeneratedCodeAttribute(\"{typeof(Emitter).Assembly.GetName().Name}\", \"{typeof(Emitter).Assembly.GetName().Version}\")]";
+
+    static readonly IndentedStringBuilder IndentedStringBuilder = new();
+
+    static void WriteNamespace(IndentedStringBuilder builder, ClassToGenerate classToGenerate)
+    {
+        if (classToGenerate.Namespace is not null)
+        {
+            builder.Append("namespace ").AppendLine(classToGenerate.Namespace)
```
MattKotsenas, line 22: I did a partial revert of the transformation of

```csharp
builder.Append("namespace ").AppendLine(classToGenerate.Namespace)
```

to

```csharp
builder.AppendLine($"namespace {classToGenerate.Namespace}")
```

(and related changes) for performance.

Because source generators must target `netstandard2.0`, the underlying `StringBuilder` doesn't have any interpolated string handling. That means interpolated strings will result in an intermediate allocation.

We _could_ write handling for interpolated strings ourselves, but the code to do so looks nontrivial: https://github.com/dotnet/runtime/pull/51653/files#diff-072e8df6b3d249d58da2b688cbd9369ae5d1e9fb4c1c1da77a19384f3f7711a0.

I also added a comment to the top of `IndentedStringBuilder` explaining the same decision.

Let me know if you strongly disagree and I can try to find another solution.

src/Verify.MSTest.SourceGenerator/Parser.cs
MattKotsenas: Now that you mention it, the type doesn't matter at all since you can't overload a property based on the return type. Removed the type check.

src/Verify.MSTest/Verifier.cs
MattKotsenas: The short answer is no. We can't use `typeof(TestContext)` because `TestContext` is an abstract base class and doesn't declare `_testMethod`. That field is declared on `Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.TestContextImplementation`, which isn't a public type in the MSTest.TestFramework package. Thus the need for the `.GetType()` call.

We _could_ make a cache of Type -> Field, however the runtime already has the `RuntimeTypeCache` (some info available @ https://learn.microsoft.com/en-us/archive/msdn-magazine/2005/july/using-net-avoid-common-performance-pitfalls-for-speedier-apps#S7 section "I Want My MemberInfo") which does it for us (verified by stepping into the framework code).

Let me know if I'm misunderstanding the request. Thanks!

src/Verify.MSTest/Verifier.cs
MattKotsenas: Same rationale as caching the field. We can't speak the `testMethod` type, hence the need for the `.GetType()` call, and the RuntimeTypeCache is caching the `AssemblyName` property lookup for us.

Let me know if I'm misunderstanding the request. Thanks!

src/Verify.MSTest/Verifier.cs
MattKotsenas: I moved the reflection work out of `Verifier` and into `TestContextReflector` to centralize all the reflection work in case we want to make additional changes now or in the future.

src/Verify.MSTest/Verifier.cs
MattKotsenas: I'm hesitant to assert on the specific shape here because I'm reaching into the internals of MSTest and not relying on any public API contracts.

More specifically, I'm concerned that this code would break and become an adoption blocker for early adopters of [TestAnywhere](https://github.com/microsoft/testfx/blob/main/docs/testingplatform/Index.md) (I'm not a user of it myself, but I did verify this code works with the current version).

In general I'm not super thrilled with the amount of reflection going on here, but I think reducing the amount would require digging more into Verify's internals, which feels to me like it should be a separate change.

Thus, I'd prefer leaving it as-is, but I'll do whatever you'd like. I'll volunteer to do follow up work in this space to reduce the amount of reflection overall, but I'd need some direction or a brain dump from you on how to proceed.

Let me know if this makes sense. Thanks!

src/Verify.MSTest.SourceGenerator/ClassToGenerate.cs
MattKotsenas: We can't rely on the autogenerated ones, as they defer equality of the collection to that type, and all the built in collection types all have reference equality semantics (even `ImmutableArray`).

At the time I wasn't sure what the source generator's data model would look like, so I brought in those two helper classes. Given that the whole data model is just `ClassToGenerate`, I implemented `Equals()` and `GetHashCode()` myself and was able to remove both classes.

src/Verify.MSTest.SourceGenerator.Tests/Tests.cs
MattKotsenas: The full incremental source generator pipeline has a good document here: https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md#caching, and this blog post is also helpful and the source of some of these test helpers: https://andrewlock.net/creating-a-source-generator-part-10-testing-your-incremental-generator-pipeline-outputs-are-cacheable/

Internally, the ValuesProvider pipeline is using both equality and hash codes. It uses the hash code first as a cheap "might be equal" check, and then if the hash codes match, uses `Equals` to perform the full equality check. https://thomaslevesque.com/2020/05/15/things-every-csharp-developer-should-know-1-hash-codes/ is a good quick reference.

MattKotsenas:
OK, all comments addressed. The two items to specifically take a look at are:

1. Take a look at the note around `IndentedStringBuilder` allocations
2. Let me know if you agree / disagree on the `TestContext` reflection choice

In all other cases I think I took your recommendations / suggestions. If you have any other feedback or thoughts, please let me know. Thanks!

src/Verify.MSTest/Verifier.cs
MattKotsenas: An idea that just hit me would be to augment what we stash in the AsyncLocal.

Rather than try to get back to the test class after-the-fact, grab everything we need in the TestContext property setter (which means we're already in the test class). Then `Verifier` or `VerifyBase` doesn't need to do any work.

If possible I'd like to keep this PR source generator focused, so if this sounds good to you I can do that as a follow up change, or even as a standalone change and then merge that into this branch.

Thoughts?

SimonCropp: > An idea that just hit me would be to augment what we stash in the AsyncLocal.

sounds good. i will merge this now. can u have a go at an AsyncLocal PR next?

PR #1204, opened 2024-05-18
Make MSTest.SourceGenerator thread-safe

MattKotsenas:
The `Execute` method on `IIncrementalSourceGenerator` is re-entrant (most easily reproducible by adding `Thread.Sleep()`s). Thus, the StringBuilder can't be static, or the competing compilations may corrupt each other's source.

Additionally, I plumbed the `CancellationToken` into more places to more closely follow the guidance from the [Incremental Generators Cookbook](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.cookbook.md).

To make the Emitter code a little easier to follow I made it an instance class. This is also what the LoggingMessage generator does (see https://github.com/dotnet/runtime/blob/af11dbc1a34538d9a38d39a7c96ddd5e7fc6907b/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen/LoggerMessageGenerator.Roslyn4.0.cs#L54-L57).


SimonCropp:
good catch.

i will deploy a beta with this merged.

PR #1205, opened 2024-05-19
Simplify reflection for MSTest's Verifier

MattKotsenas:
Previously, to get the test class's `Type`, we were either using reflection to get private properties of the `TestContext` or doing assembly scanning.

To simplify this workflow:

1. Create a new `TestExecutionContext` (open to a better name) that holds the existing `TestContext` and the required `Assembly`, `Type`, and `MethodInfo` properties
2. Update the `AsyncLocal` to hold this new `TestExecutionContext`
3. Update the source generator to populate the `TestExecutionContext` fields. At this point in the execution flow these properties are much easier to get.


src/Verify.MSTest/TestContextReflector.cs
```diff
@@ -1,129 +0,0 @@
...
-    {
-        var testName = context.TestName;
-        var typeName = context.FullyQualifiedTestClassName!; // Checked for null in the Verifier
-
-        if (!TryGetTypeFromTestContext(typeName, context, out var type))
-        {
-            type = FindType(typeName);
-        }
-
-        var testNameSpan = testName.AsSpan();
```
MattKotsenas, line 18: @SimonCropp, I don't think this code is needed anymore; let me know if you disagree.

Before this refactoring, no tests failed when I commented out this substring code. Doing a `git blame` back in time points to f4b31d2, but it's unclear to me what the previous behavior.

If you know what this code is for / is still needed let me know and I'll add a test. Thanks!

src/Verify.MSTest/TestExecutionContext.cs
```diff
@@ -0,0 +1,30 @@
+namespace VerifyMSTest;
+
+public record class TestExecutionContext
+{
+    public TestContext TestContext { get; }
+    public Assembly TestAssembly { get; }
+    public Type TestClass { get; }
+    public MethodInfo TestMethod { get; }
+
+    public TestExecutionContext(TestContext testContext, Type classType)
```
MattKotsenas, line 10: @SimonCropp open to suggestions for a name here. Honestly "TestContext" fits well, but the ambiguous type name makes things somewhat complicated.

src/Verify.MSTest/TestContextReflector.cs
SimonCropp: yeah thats a shitty change by me. lets assume for now it is no longer required.  i will advertise the beta and hopefully if it is required, it will become apparent why

src/Verify.MSTest/TestExecutionContext.cs
SimonCropp: happy with your choice

src/Verify.MSTest/TestExecutionContext.cs
```diff
@@ -0,0 +1,20 @@
+namespace VerifyMSTest;
+
+public record TestExecutionContext(TestContext TestContext, Type TestClass)
+{
+    public Assembly TestAssembly { get; } = TestClass.Assembly;
```
SimonCropp, line 5: lets go with `Aeembly` and `Method as property names. the `Test` part is assumed

src/Verify.MSTest/TestExecutionContext.cs
```diff
@@ -0,0 +1,20 @@
+namespace VerifyMSTest;
+
+public record TestExecutionContext(TestContext TestContext, Type TestClass)
```
SimonCropp, line 3: can this be internal?

PR #1207, opened 2024-05-21
Don't skip MSTest source generator when base has manually defined TestContext

MattKotsenas:
Based on using the beta MSTest source generator, I'm removing the skip of emitting the `TestContext` property when the base class has a manually defined property.

Before this change, if there were duplicate definitions of `TestContext`, the manual one would win, however the manual one _probably_ doesn't set up the `TestExecutionContext` and thus the test would fail at runtime. After this test the compiler throws the standard "CS0108: member1' hides inherited member 'member2'. Use the new keyword if hiding was intended" message, which makes it much more obvious what the generator is doing and how to fix it.

As part of the change I also split the singular `Tests` class into multiple classes, added more tests, and moved some internal classes out of the <global> namespace and into `VerifyMSTest.SourceGenerator`.


src/Verify.MSTest.SourceGenerator/IndentedStringBuilder.cs
```diff
@@ -1,3 +1,5 @@
+namespace VerifyMSTest.SourceGenerator;
```
SimonCropp, line 1: is the namespace required?

would seem not for an internal type

SimonCropp:
i will deploy this now

PR #1209, opened 2024-05-22
Support [UsesVerify] as assembly attribute in MSTest

MattKotsenas:
Refactor the incremental generator pipeline so that `[UsesVerify]` can be either on a class or on the assembly. When on a class, the class is a candidate for generation, when on the assembly, all classes with the `[TestClass]` attribute are candidates for generation.

Now the pipeline starts as two pipelines, the existing `ForAttributeWithMetadataName` pipeline and a new `SyntaxProvider` pipeline that early in the transform checks the assembly for the attribute, then follows a similar pattern. The assembly attribute path is slightly more complicated because `[TestClass]` isn't a sealed attribute (because it's an extensibility point in MSTest), so we need to search for an attribute of type `TestClassAttribute` or a type that derives from `TestClassAttribute`. Once each pipeline is complete, there's a `.Combine().SelectMany()` step to essentially flatten the two collections into a single collection.

As part of this change I did a fair amount of refactoring. There's not much logic change, mostly two types of refactorings:

1. Favoring multiple smaller extension methods over a few longer functions
2. Simplifying the role of `Parser` - now it's only responsible for collecting the information necessary to build `ClassToGenerate` and any candidate filtering logic is moved into the generator's main transform method

I did some additional fixes that I saw while writing this PR. I can back them out and submit them separately if desired.

1. Added more unit tests
2. Added the SourceGenerator to AppVeyor
3. Learned that the `[GeneratedCode]` attribute should be on the generated property, not the class (otherwise it applies to the user's portion of the partial class as well, see https://learn.microsoft.com/en-us/archive/blogs/codeanalysis/correct-usage-of-the-compilergeneratedattribute-and-the-generatedcodeattribute)


MattKotsenas:
@SimonCropp I think after this we're ready for feedback from the community. 

Let me know if you have any questions!

SimonCropp:
awesome. thanks

PR #1224, opened 2024-06-12
file nesting props include (

zachdean:
change nuget targets to allow file nesting with nunit param test fixtures

Before:
image

After:
image


zachdean:
@dotnet-policy-service agree

SimonCropp:
@zachdean thanks

PR #1237, opened 2024-07-02
UseStrictJson should result in json extension

SimonCropp:


304NotModified:
Thanks, after this change every works well :)

PR #1241, opened 2024-07-14
Show how to ignore class level params in NUnit tests

tom-englert:
IgnoreParametersForVerified only works on method params, not class params.

This PR adds tests that show a workaround, using a combination of UseTypeName and DisableRequireUniquePrefix, to remove the class params from the output file name.


SimonCropp:
@tom-englert yeah that is a poor implementation on my part. you should not need to hack around it

thoughts on this as a better fix https://github.com/VerifyTests/Verify/pull/1249  ?

tom-englert:
This workaround was fine for my simple use case, however #1249 would be more comprehensive.

SimonCropp:
deploying v26 now

PR #1249, opened 2024-07-17
Breaking: Respect NUnit class level parameters and drop support for custom TestName

SimonCropp:
## Who is effected

 * NUnit users who are using a parameterised `TestFixture`
 * NUnit users who are using `TestCase.TestName`

### No longer respect `TestFixture.TestName` and `TestCase.TestName

Reasons

 * NUnit does not enforce uniqueness for these names. Which can result in difficult to diagnose file conflicts in received and verified files.
 * Prevents received and verified files for a specific test class or test method from being group together.
 * NUnit provides no way to reliably get TestName, resulting in Verify needing to guess the name using some string manipulation

### Respect fixture level parameters and ignore NUnit fixture name convention

Previoulsy fixture level parameters were properly not respected by Verify. Verify would rely on NUnit convention for the fixture name. For example

```
[TestFixture("2", TestName = "CustomClassName")]
public class ClassLevelParams(string classParam)
```

The NUnit fixture name would be `ClassLevelParams(1)`. While unique enough for the verified file it has several problems

 * Does not convey the parameter name in the resulting file
 * Make the Handling of `TestFixture` parameters inconsistent with `TestCase` parameters
 * Does not match the Verify convention of `paramName=paramValue`
 * Is incompatible with several Verify features. For example [Ignore parameters for verified filename](https://github.com/VerifyTests/Verify/blob/main/docs/parameterised.md#ignore-parameters-for-verified-filename) and [NameForParameter](https://github.com/VerifyTests/Verify/blob/main/docs/parameterised.md#complex-memberdata)

## Sample

```
[TestFixture("1")]
[TestFixture("2", TestName = "CustomClassName")]
public class ClassLevelParams(string classParam)
{
    [TestCase("3")]
    [TestCase("4", TestName = "CustomClassName")]
    public Task Test(string methodParam) =>
        Verify("Value");
```

### Before

```
ClassLevelParams(1).CustomClassName_methodParam=4.verified.txt
ClassLevelParams(1).Test_methodParam=3.verified.txt
CustomClassName.CustomClassName_methodParam=4.verified.txt
CustomClassName.Test_methodParam=3.verified.txt
```

### After

```
ClassLevelParams.Test_classParam=1_methodParam=3.verified.txt
ClassLevelParams.Test_classParam=1_methodParam=4.verified.txt
ClassLevelParams.Test_classParam=2_methodParam=3.verified.txt
ClassLevelParams.Test_classParam=2_methodParam=4.verified.txt
```

## Migration

After upgrading, the recommended approach is to treat the old verified files as [Orphaned verified files](https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#orphaned-verified-files).


tom-englert:
👍 ❤️

PR #1267, opened 2024-08-12
Faster locking through .NET 9's System.Threading.Lock

MarkCiliaVincenti:


SimonCropp:
can you quantify "faster"

MarkCiliaVincenti:
> can you quantify "faster"

https://steven-giesel.com/blogPost/4cf6d48a-ec9d-4c68-961c-31fd8d8c1340

MarkCiliaVincenti:
Basically System.Threading.Lock on .NET 9.0 is benchmarked to be around 25% faster. This PR brings this performance improvement into existing locks for .NET 9+ projects depending on this library.

SimonCropp:
given locking is only used in recording, not a hot path and a rarely used api, it is not worth the extra reference

MarkCiliaVincenti:
@SimonCropp a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

PR #1291, opened 2024-09-12
Fix #1290: Ignore only class level parameters in NUnit tests is not possible

### LINKED ISSUE #1290: Ignore only class level parameters in NUnit tests is not possible
#### Describe the bug

It's not possible to ignore only the class level parameters of an NUnit test, but include the method level parameters.

#### Minimal Repro

These are the combinations I tried:

```c#
[TestFixture("1")]
[TestFixture("2")]
public class ClassLevelParams2(string arg1)
{
    [TestCase("3")]
    [TestCase("4")]
    public Task IgnoreClassLevelParametersOnly_Attempt1(string arg2) =>
        Verify(arg2)
            .IgnoreParametersForVerified(arg1);

    [TestCase("3")]
    [TestCase("4")]
    public Task IgnoreClassLevelParametersOnly_Attempt2(string arg2) =>
        Verify(arg2)
            .IgnoreParametersForVerified(arg2);

    [TestCase("3")]
    [TestCase("4")]
    public Task IgnoreClassLevelParametersOnly_Attempt3(string arg2) =>
        Verify(arg2)
            .UseParameters(arg2);

    [TestCase("3")]
    [TestCase("4")]
    public Task IgnoreClassLevelParametersOnly_Attempt4(string arg2) =>
        Verify(arg2)
            .UseParameters(arg2)
            .DisableRequireUniquePrefix();

}
```

These are the files created:
```
ClassLevelParams2.IgnoreClassLevelParametersOnly_Attempt1.verified.txt
ClassLevelParams2.IgnoreClassLevelParametersOnly_Attempt2.verified.txt
ClassLevelParams2.IgnoreClassLevelParametersOnly_Attempt3_arg1=3.verified.txt
ClassLevelParams2.IgnoreClassLevelParametersOnly_Attempt3_arg1=4.verified.txt
ClassLevelParams2.IgnoreClassLevelParametersOnly_Attempt4_arg1=3.verified.txt
ClassLevelParams2.IgnoreClassLevelParametersOnly_Attempt4_arg1=4.verified.txt
```

Attempt 1 & 2 fail because no parameters are used at all.
Attempt 3 fails partially with " System.Exception : The prefix has already been used"
Attempt 4 works, however like in 3 the name of the parameter is wrong, should be `arg2`

tom-englert:
This will introduce a new setting `IgnoreParameters`, the parameters of this setting are the **names** of the parameters to be ignored.

The method `UseParameters` is expected to always contain all parameters, even if some of them will be ignored!
Also `UseParameters` needs to be called along with `IgnoreParameters` for test frameworks that do not support automatic parameter detection like NUnit.


SimonCropp:
> Since the parameters passed to IgnoreParametersForVerified had no meaning/effect before (always all parameters were ignored), this should be an acceptable change.

i dont think this is true. the parameters passed in are used to prevent conflicts in received files

tom-englert:
I see, maybe this was only true for NUnit, because there UseParameters is optional...
Let me rethink this

tom-englert:
OK, introduced a new setting, so it does not collide with the original one

SimonCropp:
thanks. will include in the next release

PR #1307, opened 2024-09-29
Fix StaticAutoVerifyDelegate example

jerone:
The globally settings delegate is using the wrong example: https://github.com/VerifyTests/Verify/blob/main/docs/verify-options.md#globally


SimonCropp:
@jerone thanks

PR #1327, opened 2024-10-28
Builder Pattern - Allow chaining settings configuration calls

thomhurst:
Feel free to close if you don't want this change.

If you have various options you want to set, it can feel verbose to have to declare a variable and then repetitively having to access it and call various methods on it.

This change would introduce the builder pattern to `VerifySettings` so that methods that amend its configuration return you back the same object, allowing you to chain settings easily and concisely.

This means you don't necessarily have to declare a variable either, you could just inline the settings object into your `Verify(...)` call.

Take this from your documentation:

```csharp
[Fact]
public Task OnCallbacks()
{
    settings.OnVerify(
        before: () => Debug.WriteLine("before"),
        after: () => Debug.WriteLine("after"));
    settings.OnFirstVerify(
        (receivedFile, receivedText, autoVerify) =>
        {
            Debug.WriteLine(receivedFile);
            Debug.WriteLine(receivedText);
            return Task.CompletedTask;
        });
    settings.OnVerifyMismatch(
        (filePair, message, autoVerify) =>
        {
            Debug.WriteLine(filePair.ReceivedPath);
            Debug.WriteLine(filePair.VerifiedPath);
            Debug.WriteLine(message);
            return Task.CompletedTask;
        });

    return Verify("value");
}
```

This could become:

```csharp
[Fact]
public Task OnCallbacks() {
    return Verify("value", new VerifySettings()
        .OnVerify(
            before: () => Debug.WriteLine("before"),
            after: () => Debug.WriteLine("after"))
        .OnFirstVerify(
            (receivedFile, receivedText, autoVerify) => {
                Debug.WriteLine(receivedFile);
                Debug.WriteLine(receivedText);
                return Task.CompletedTask;
            })
        .OnVerifyMismatch(
            (filePair, message, autoVerify) => {
                Debug.WriteLine(filePair.ReceivedPath);
                Debug.WriteLine(filePair.VerifiedPath);
                Debug.WriteLine(message);
                return Task.CompletedTask;
            }));
}
```

This of course doesn't stop you defining it the old way either - So the choice is up to the user and remains backwards compatible.

### DIFF
--- a/src/Verify.Expecto.FSharpTests/Tests.fs
+++ b/src/Verify.Expecto.FSharpTests/Tests.fs
@@ -18,7 +18,7 @@ let tests =
 let uniqueTests =
     testTask "unique" {
         let settings = VerifySettings()
-        settings.UniqueForRuntime()
+                           .UniqueForRuntime()
         do! Verifier.Verify("unique", "value", settings)
     }
 // end-snippet
@@ -27,7 +27,7 @@ let uniqueTests =
 let typeNameTests =
     testTask "typeNameTests" {
         let settings = VerifySettings()
-        settings.UseTypeName("CustomTypeName")
+                           .UseTypeName("CustomTypeName")
         do! Verifier.Verify("typeNameTests", "Value", settings)
     }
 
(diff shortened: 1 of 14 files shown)


SimonCropp:
there is already a pattern for fluent API. so i added callbacks to that API: https://github.com/VerifyTests/Verify/pull/1332

PR #1336, opened 2024-10-31
Fix TUnit Sample

thomhurst:
Looks like a minor copy+paste error. TUnit doesn't have a `[TestFixture]` attribute 😄


SimonCropp:
thanks

PR #1356, opened 2024-11-19
feat: Scrub/ignore by predicate function

JulianRooze:
I'm trying to introduce Verify to our large existing code base and one area of friction is that we have thousands upon thousands of dynamic properties (auto-generated IDs) that we want to ignore or scrub, which all follow a similar pattern, they all end in `ID`. While it's possible to use reflection at startup to find all those properties and Scrub/Ignore them one by one, it's a bit wasteful as only a small subset of them will ever pass through a call to `Verify`, we have to pre-emptively ignore them ahead of time. For this, I think the ability to provide a callback function that gets called at the same point the normal "is property ignore/scrub dictionary" is a good approach. The callback can be based on either the string name of the property (or anything that's not a property like a dictionary key) or on the `MemberInfo` of the property if you need more information to determine if the property should be ignored or not. 

I did not include benchmarks as I hope from the implementation it should be clear that there shouldn't be any performance impact from this, especially not if you don't opt-in to this feature, but let me know if you must have them. 

Todo:
- Docs I guess?

This is based on the discussion here:

https://github.com/orgs/VerifyTests/discussions/1345


JulianRooze:
@dotnet-policy-service agree [company="New Black"]

SimonCropp:
yep will need some docs.

can u also add static settings

JulianRooze:
@SimonCropp is that on `public static partial class VerifierSettings`? If so, those are included [here](src/Verify/Serialization/VerifierSettings_SerializationMaps.cs). If not, then I'm afraid I need a pointer where exactly. 

As for docs: happy to write some examples, but there's some auto-generation of docs going on and I couldn't get it to work. 

What I tried was adding a `#region ScrubMemberByPredicate` around my tests and then adding this to the docs:

```
## Scrub members by predicate

<!-- snippet: ScrubMemberByPredicate -->
<!-- endSnippet -->
```

But if I then build the solution, it simply removes that section from the docs. What exactly is the flow to add docs?

benmccallum:
@JulianRooze , it would be using https://github.com/SimonCropp/MarkdownSnippets, but I imagine the doc generation is handled by the workflow [here](https://github.com/VerifyTests/Verify/blob/main/.github/workflows/on-push-do-docs.yml)

PR #1362, opened 2024-11-28
Add support for TUnit Class Constructors

SimonCropp:
https://github.com/orgs/VerifyTests/discussions/1360


SimonCropp:
i would have though the file name should be the combination of the args from the constructor and the args from the test method?

Issue #1368, opened 2024-12-11
Incorrect scrubbing of DateTime


greri-tricentis:
When the culture is set to `en-US`

and I do 

```
[Test]
public Task TestScrubberOfDates() {
    var settings = new VerifySettings();
    settings.ScrubInlineDateTimes("G");
    return Verify("12/11/2024 10:36:43 AM", settings);
}
```

then I receive
```
1DateTime_1
```

I did however expect it to yield
```
DateTime_1
```

This is because the "Long" Date in this Class has a 9
https://github.com/VerifyTests/Verify/blob/a148ff11ea98e12d52e277b547ae0acd55905e89/src/Verify/Serialization/Scrubbers/DateScrubber_Generated.cs#L4
But a date with a 9 is not as long as it could be. So it will not recognize `12/11/2024 10:36:43 AM` as a datetime as its too long.
It will however find `2/11/2024 10:36:43 AM` and scrub that.

SimonCropp:
can u submit a failing unit test as a pull request

SimonCropp:
can u try version 28.7.1

greri-tricentis:
it works

PR #1377, opened 2024-12-21
Incorrect scrubbing of DateTime - Unit test

mithileshz:
This PR is adding the unit test for #1368 and maybe the potential fix?

The issue is that when the culture is set to `en`, the `long` date in `src/Verify/Serialization/Scrubbers/DateScrubber_Generated.cs` contains one character less than the possible max (the month is 9 instead of 10, 11, or 12). This means that the scrubber does not scrub the first character and therefore returns `1DateTime_1` in the received file. 

I was not sure how `src/Verify/Serialization/Scrubbers/DateScrubber_Generated.cs` was generated so this change was done manually, however I think this change might need to be applied across all the different cultures?


gregorriegler:
oh thank you. I was just forking to create this. Now I see that you already did ☺️

SimonCropp:
can u try version 28.7.1

Issue #1383, opened 2025-01-09
Missing F# `None` field in the generated string


MangelMaxime:
#### Describe the bug

I am trying to setup Verify with xUnit v3 for F#, but it seems like `None` field are not included in the output.

#### Minimal Repro

```fs
module Ionide.XmlDocFormatter

open Xunit
open System.Threading.Tasks
open VerifyXunit
open VerifyTests
open Argon
open System

// I need to use Fixture in order for the settings to be applied
type SetupFixture() =
    do
        // Support F# types
        VerifierSettings.AddExtraSettings(fun settings ->
            settings.AddFSharpConverters()
            settings.NullValueHandling <- NullValueHandling.Include
        )

        // Put Verify files in a separate directory
        Verifier.UseProjectRelativeDirectory("VerifyTest")

    interface IDisposable with

        member __.Dispose() = ()

type Tests() =
    interface IClassFixture<SetupFixture>

    [<Fact>]
    member __.``Can create a start node`` () : Task =

        Verifier.Verify({|
            Name = "Maxime"
            Age = Some 12
            Street = None // This field is missing in the generated result
        |}).ToTask()
```

I get this output, but I think it should include the `Street` property as I asked to include `null`. [Documentation](https://github.com/VerifyTests/Verify/blob/main/docs/fsharp.md#nullvaluehandling)

```json
{
  Age: {
    Case: Some,
    Fields: [
      12
    ]
  },
  Name: Maxime
}
```

1. Download [Ionide.XmlDocFormatter.zip](https://github.com/user-attachments/files/18363955/Ionide.XmlDocFormatter.zip)
2. `dotnet tool restore`
3. `cd tests`
4. `dotnet test`

SimonCropp:
i think that documentaiton might be incorrect. can u try setting `settings.DefaultValueHandling` to `DefaultValueHandling.Include`

MangelMaxime:
Using `settings.DefaultValueHandling <- DefaultValueHandling.Include` seems to be correct:

```text
{
  Age: {
    Case: Some,
    Fields: [
      12
    ]
  },
  Name: Maxime,
  Street: null
}
```

PR #1385, opened 2025-01-10
Update F# documentation on including `None`

MangelMaxime:
Related to #1383


MangelMaxime:
@dotnet-policy-service agree

SimonCropp:
thanks. and sorry for the inconvenience

MangelMaxime:
No worries, I didn't really need F# type support at the time only string but I though why not configured everything just in case.

Thanks for this library it makes testing easier to deal with especially for text.

PR #1394, opened 2025-02-03
Fix for TUnit v0.10 Signature Changes

thomhurst:
Apologies but I did some refactoring and signature changes in the latest TUnit. This fixes references so it's compatible for the new version.


SimonCropp:
> Apologies but I did some refactoring and signature changes in the latest

no worries. and thanks for the fix. i will deploy this now

PR #1400, opened 2025-02-14
Added a readme file to the NuGet packages

Guiorgy:
With the Visual Studio 17.13 update the NuGet package manager/browser [shows a preview of the included readme](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes#net).

Might want to consider embedding a readme with the packages moving forward, otherwise Visual Studio shows the following:

missing readme preview in visual studio

Ideally, each package would have a dedicated readme, but that would probably be too annoying to maintain, so for now I included the same file in every package.

I included the `readme.source.md` file, though if you'd prefer `readme.md` instead, I can change that.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -8,6 +8,7 @@
     <AssemblyVersion>1.0.0</AssemblyVersion>
     <PackageTags>Json, Testing, Verify, Snapshot, Approvals</PackageTags>
     <Description>Enables verification of complex models and documents.</Description>
+    <PackageReadmeFile>readme.source.md</PackageReadmeFile>
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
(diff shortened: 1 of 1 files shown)


Guiorgy:
@dotnet-policy-service agree

SimonCropp:
doesnt this make it difficult https://github.com/NuGet/NuGetGallery/issues/8644  ?

SimonCropp:
also Mermaid  diagrams are not supported https://github.com/NuGet/NuGetGallery/issues/8942

Guiorgy:
Good point

SimonCropp:
so whats the way forward?

Guiorgy:
There are several options:

- Keep things as they were
- Embed one of the readmes as they are ignoring the problems
- Make a smaller simplified readme for embedding into the NuGet packages

Your choice really.

SimonCropp:
> Make a smaller simplified readme for embedding into the NuGet packages

happy to consider a PR that does this

PR #1426, opened 2025-03-16
Add .NET Framework support to Verify.TUnit

thomhurst:


SimonCropp:
but why? you cant run tests on netstandard

thomhurst:
> but why? you cant run tests on netstandard

So they it can be used in .NET framework projects, but I see you're just directly referencing .NET framework for the other testing libs

SimonCropp:
ahh. yep net classic support is a better approach

thomhurst:
Switched to framework :smile:

SimonCropp:
deployed. give nuget approx 10min

PR #1449, opened 2025-04-29
remove support for hashing parameters

SimonCropp:
ie the `HashParameters` method on settings.

It is getting difficult to maintain this since [System.IO.Hashing](https://www.nuget.org/packages/System.IO.Hashing/) has dropped support for net6 and net7

also IMO i think it is a flawed feature. and instead for long parameter names, they should be overridden on a per-parameter basis using https://github.com/VerifyTests/Verify/blob/bed799382b1c296374416ca518a6c5e2fdb9468b/docs/parameterised-nunit.md#useparameters

https://github.com/VerifyTests/Verify/blob/bed799382b1c296374416ca518a6c5e2fdb9468b/docs/parameterised-nunit.md#hashing-parameters


michaelmadegard:
Hi @SimonCropp !
Did you put this feature in a separate nuget ?
I was using this very helpful feature as I was experiencing issues with the Windows "paths too long" thing. It's a pain it just disappeared with the update of my version of Verify.
How can the "UseTextForParameters" be used to achieve the same behaviour ?
Thanks !

SimonCropp:
@michaelmadegard can you share an example of you test and the parameter values

michaelmadegard:
> @michaelmadegard can you share an example of you test and the parameter values

Actually I'm using Verify alongside SpecFlow so the underlying code is generated. But these steps in SpecFlow simply verify the content of ZIP archives that are generated by my code, such as :

`await VerifyZip(_reportContext.ReportPath).HashParameters();`

I do not provide anything but the path to the ZIP archive.
The `HashParameters` fluent method was helping me shortening the name of the generated ".verified" file.

SimonCropp:
example of your .verified file name?

michaelmadegard:
Actually my case relates to the name of the folder that contains the .verified files, as it concerns an archive that is unzipped.
With the HashParameters the folder is :

`RemainingCapitalFeature.Test2_db93338260f1e638`

Without the HashParameters it became something like :

`RemainingCapitalFeature.Test2_myReference=ABC20250507ABCDEFG_folder=Test_remainingAmount=1234.56_status=PERF_exampleTags=null`

The .verified files inside the unzipped folder are OK.

SimonCropp:
and the combination of all those parameters 5 parameters is distinct for every test ?

michaelmadegard:
Yes, so I correctly got one folder (with a unique hash) by test case.

samtrion:
@SimonCropp I also use the `HashParameters()` functionality extensively in some projects.

If this is omitted, I sometimes have file names that are approx. 400-500 characters long.

Would it be an option to work with a conditional versioning of `System.IO.Hashing` here? Something like the following?

```xml
<ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net7.0' ">
  <PackageVersion Include="System.IO.Hashing" Version="8.0.0" />
</ItemGroup>
<ItemGroup Condition=" '$(TargetFramework)' != 'net6.0' and '$(TargetFramework)' != 'net7.0' ">
  <PackageVersion Include="System.IO.Hashing" Version="9.0.4" />
</ItemGroup>
```

SimonCropp:
@michaelmadegard @samtrion  can u try this https://github.com/VerifyTests/Verify.ParametersHashing

samtrion:
> @michaelmadegard @samtrion can u try this https://github.com/VerifyTests/Verify.ParametersHashing

Looks good, perfect 1:1 replacement without any code changes after upgrading to 30.0.0 and additional package `Verify.ParametersHashing`

michaelmadegard:
> @michaelmadegard @samtrion can u try this https://github.com/VerifyTests/Verify.ParametersHashing

Hello @SimonCropp,
Just like @samtrion, everything goes fine with the new package added to the last version of Verify.
Thank you for your kind help !

PR #1470, opened 2025-06-08
feat: Add option to disable adding received files as attachments

MJB222398:
This isn't suitable for approval, my new tests have made existing ones flakey. ScrubberLevelsSample tests often fail, I think maybe if my tests are running and reset the VerifierSettings so they lose the scrubber? I can't seem to get it working. So seeking some guidance. Do you think the approach in general is correct here? Are you able to provide some input on how best to get these tests passing?


SimonCropp:
i think the only way to get it to work is have those tests in their own test projects

MJB222398:
Thanks that seems to work. This build has failed here though:

The active test run was aborted. Reason: Test host process crashed : Process is terminated due to StackOverflowException.

I don't see this locally.

src/Verify.MSTest.DisableAttachments.Tests/DisableAttachmentsTests.cs
```diff
@@ -0,0 +1,24 @@
...
+        ResultFilesCallback.Callback = list => Assert.AreEqual(0, list.Count);
+        VerifierSettings.DisableAttachments();
+        var settings = new VerifySettings();
+        settings.DisableDiff();
+        await Assert.ThrowsExceptionAsync<VerifyException>(
+            () => Verify("Bar", settings));
+    }
+
+    [TestCleanup]
+    public void ResetStaticSettings()
```
SimonCropp, line 19: given there is only one test. isnt this redundant?

SimonCropp:
thanks. i will deploy this now

MJB222398:
Nice, thanks!

Issue #1473, opened 2025-06-24
Documentation on VerifySettings.UseParameters is incorrect for MSTest


ProbablePrime:
#### Describe the bug

In, the reference documentation within the source for [`VerifySettings.UseParameters();`](https://github.com/VerifyTests/Verify/blob/main/src/Verify/Naming/VerifySettings_Parameter.cs#L45) It states that `UseParameters()` is required for MSTest.

However, the [Readme documentation for this](https://github.com/VerifyTests/Verify/blob/main/docs/parameterised-mstest.md#useparameters) states that UseParameters is not required.

The documentation should agree with each other. I think the right move here is to update the docs in the source to match the readme.

SimonCropp:
happy to consider a pull request

ProbablePrime:
Of course, I will submit this soon.

ProbablePrime:
#1476 is merged so closing this.

PR #1476, opened 2025-06-30
docs(verifysettings): correct UseParameters docs for mstest

ProbablePrime:
As discussed in: https://github.com/VerifyTests/Verify/issues/1473


SimonCropp:
thanks

PR #1480, opened 2025-07-11
Test Projects: Include Verify.props and Verify.targets in sequence similar to nuget import sequence

BrunoJuchli:
By importing them via convention via Directory.Build.props and Directory.Target.props

image

> [!Note]
> DeterministicTests.csproj still explicitly imports the files. This is because it defines its own Directory.Build.props, overriding the src directory's Directory.Build.props (not exactly sure if this was intended or not...).


SimonCropp:
@BrunoJuchli very nice. thanks

Issue #1478, opened 2025-07-02
Cannot override `<DependentUpon>` metadata for `*.verified.*` files.


BrunoJuchli:
## TL;DR @SimonCropp 
Is there a specific reason the `<None Update>` msbuild `<ItemGroup>` elements are part of the **.targets** file and not the **.props* file?

Thank you 👍 

## Symptom

I cannot override the `<DependentUpon>` msbuild metadata for my *.verified.* files.

#### Minimal Repro

https://github.com/BrunoJuchli/VerifyOverridesDependentUponRepro

> [!Note]
> I have asserted that the issue only occurs when i include a package reference  to `Verify`. As soon as I remove the reference, the issue vanishes.

I have a folder structure like (*1):
- SomeFoo.cs
- SomeFoo{TOne}.verified.csv
- SomeFoo{TTwo}.verified.csv

Now I tried to put these files in the visualization under SomeFoo.cs, like:

- SomeFoo.cs
  - SomeFoo{TOne}.verified.csv
  - SomeFoo{TTwo}.verified.csv

It doesn't work out of the box, because the Verify.targets sets `<DependentUpon>` to `%(ParentFile).cs` which evaluates to `SomeFoo{TOne}`.

So I tried overriding `<DependentUpon>`. No matter what I set in my csproj, it doesn't have any effect.
I also tried overriding `<ParentFile>`, but that also doesn't work.

I have set other metadata names and I get the expected result, for example
```xml
<MyProperty>$([System.Text.RegularExpressions.Regex]::Match(%(Filename)%(Extension), `^[^{\\.]*`)).cs</MyProperty>
```

## Problem Analysis

Verify sets both `<ParentFile>` and `<DependentUpon>` in it's Verify.**targets** file.
This is evaluated after my own content in my csproj.
So it overrides what ever i set.

## Solution Proposal

Move the `<None Update>` elements to Verify.**props**.

I can create a PR for this, but I'm not sure if moving it to the props files would cause other issues. So let me ask first:
**Is there a specific reason the `<None Update>` elements are part of the **.targets** file and not the **.props* file?

Possibly we'll have to change from `<None Update>` to `<None Include>`, if the default imports of .net sdk / c# for these files come after the import of Verify.props.

### Alternative
Another `..Split('{')[0]` could be added to the pattern.
I don't think this is great. It only solves my problem. And I could just as well rename my files to use `()` instead of `{}`...
So this is not really a serious suggestion. In fact, I would advise against it.

[...]

PR #1481, opened 2025-07-11
#1478: Allow Overriding of DependentUpon by moving from .targets to .props

BrunoJuchli:
This PR addresses #1478

Excerpt from [Microsoft.NET.Sdk.props](https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props):

```xml
  <!--
      Extensibility point for users to import build logic after all of the .NET SDKs .props are imported but before the project's contents are evaluated.
      This can be useful for overriding default globs or properties that the SDK sets. 
      -->
  <Import Project="$(AfterMicrosoftNetSdkProps)" Condition="$(AfterMicrosoftNetSdkProps) != ''" />
```

To achieve this, I have:
- moved specifying of DependentUpon to a new `Verify.AfterMicrosoftNetSdk.props` file.
- adapted `Verify.props` to specify 
  `<AfterMicrosoftNetSdkProps>$(AfterMicrosoftNetSdkProps);$(MSBuildThisFileDirectory)Verify.AfterMicrosoftNetSdk.props</AfterMicrosoftNetSdkProps>`
- remove Verify.targets - it's empty now.

Also: changed form `None Update` to `None Include`, This to prevent that opting out of the default includes of .net sdk to also affect Verify's includes.
I'm not sure what exactly the desired behavior is. I noticed that previously `Verify.targets` used `None Update` for C# and VB, but `None Include` for F#. I suspect this was not intended.

### Behavior Options

- use `None Include`. The files are always included unless the user adds `None Remove` to remove them again.
- use `None Update` -- user can control inclusion via `EnableDefaultItems` and `EnableDefaultNoneItems` (flags defined in `Microsoft.NET.Sdk.DefaultItems.props`.
- add our own flag to control inclusion `EnableVerifyDefaultItems`
  - and use `None Update` --> inclusion can be switched off by any off `EnableVerifyDefaultItems`, `EnableDefaultItems` or `EnableDefaultNoneItems`, **or**
  - and with `None Include` --> inclusion exclusively controlled by `EnableVerifyDefaultItems`, **or**
  - we control exclusively by `EnableVerifyDefaultItems` and have a condition so that:
    - if `EnableDefaultItems` or `EnableDefaultNoneItems` are false: use `None Include`
    - otherwise: use `None Update`.

> [!Note]
> using `None Include` seems to override previous include's metadata. That's why having our include first didn't work.

### DIFF
--- a/src/Verify.Tests/NugetTests.Run/structure.verified.md
+++ b/src/Verify.Tests/NugetTests.Run/structure.verified.md
@@ -23,8 +23,10 @@
 * icon.png
 * buildTransitive
   * Verify.props
+  * Verify.AfterMicrosoftNetSdk.props
 * build
   * Verify.props
+  * Verify.AfterMicrosoftNetSdk.props
 * [Content_Types].xml
 * package
   * services
--- a/src/Verify/Verify.csproj
+++ b/src/Verify/Verify.csproj
@@ -15,6 +15,8 @@
 
     <None Include="buildTransitive\Verify.props" Pack="true" PackagePath="buildTransitive\" />
     <None Include="buildTransitive\Verify.props" Pack="true" PackagePath="build\" />
+    <None Include="buildTransitive\Verify.AfterMicrosoftNetSdk.props" Pack="true" PackagePath="buildTransitive\" />
+    <None Include="buildTransitive\Verify.AfterMicrosoftNetSdk.props" Pack="true" PackagePath="build\" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
--- a/src/Verify/buildTransitive/Verify.AfterMicrosoftNetSdk.props
+++ b/src/Verify/buildTransitive/Verify.AfterMicrosoftNetSdk.props
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <None Include="**\*.received.*;**\*.verified.*" Condition="$(Language) == 'C#'">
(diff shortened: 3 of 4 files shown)


BrunoJuchli:
Update:
I think Rider diagnostics display is just wrong.
When looking at the binary ms build log, it shows that the files are included twice as `None`. The first include contains the metadata as set by Verify.props, the second has no metadata (as included from .net sdk).
So it looks like the second entry simply replaces the first entry (I was misslead here by AI which told me it would merge the metadata.... which is what Rider seems to be doing?!).

-----

@SimonCropp 
This is a bit of a head scratcher for me: even after performing `git clean -xdf` and restarting rider, rider doesn't show the verified.txt files nested under the expected node:
image

But Riders diagnostic output gives me all the correct data to indicate it should be nested under `ComplexParametersSample.cs`:
image

Could you check out this branch and try it out on your machine, please? Thanks.

Edit: I can reproduce the same behavior in VSCode. So maybe the diagnostic output of rider is wrong.
Edit 2: I tried generating a binlog by building on the cmd line, but I get errors like `InternalsVisibleToAttribute` could not be found. Even though I chose msbuild.dll of 10.0.100-preview 5 in Msbuild structured log viewer. Maybe a problem with the viewer...

BrunoJuchli:
The good news is, that overriding in the csproj now works... ^^

image

BrunoJuchli:
@SimonCropp it works now 👍 Tested in VSCode and Rider.

Rider still isn't showing correct diagnostic info. It looks like it just picks the first match - which is not how msbuild itself evaluates it...
(I've created an issue for Rider: https://youtrack.jetbrains.com/issue/RIDER-127730)

SimonCropp:
i tested this, and yes it does seem to work. I will deploy it now

sorry for the delayed response

PR #1487, opened 2025-07-26
Add two YellowDogMan Extensions

ProbablePrime:
Adds the two extensions I recently wrote that are over on YellowDogMan.


SimonCropp:
awesome. thanks

Issue #1482, opened 2025-07-16
Naming Conflict TestContext with Playwright.NET


msexxeta:
#### Describe the bug

There is a conflict with Playwright.NET and Verify.MSTest in that it already defines a "TestContext" so marking a class that uses playwright with UsingVerify produces a compile error:

> 'MyPageTests.TestContext': cannot override inherited member 'WorkerAwareTest.TestContext' because it is not marked virtual, abstract, or override

This could of course also be solved in the Playwright.NET repo but I think it's unlikely that they will rename their context.

#### Minimal Repro

```cs
[TestClass]
[UsesVerify]
public partial class MyPageTests : PageTest
{
}
```

#### Submit a PR that fixes the bug

I have not made a PR, but renaming the property TestContext would solve this issue. Maybe keep the name as TestContext as default, but make UsesVerifyAttribute take options where you could specify a name like
```cs
[UsesVerify(testContext: "VerifyContext")]
```

SimonCropp:
@MattKotsenas thoughts on this one?

MattKotsenas:
Seems reasonable. Before saying any more, we need to do two simple experiments:

1. Does MSTest still property inject the context if it has a different name (i.e. does it inject by name or by type)?
2. If yes, will it inject the property _multiple_ times, or is the Playwright property a different type?

MattKotsenas:
So it looks like the answer to both 1 and 2 is "no". When I use this simple repro:

```csharp
[TestClass]
public sealed class Test1
{
    [TestMethod]
    public void TestMethod1()
    {
        Assert.IsNotNull(TestContext, $"{nameof(TestContext)} should not be null");
        Assert.IsNotNull(VerifyTestContext, $"{nameof(VerifyTestContext)} should not be null"); // fails!
    }

    public TestContext TestContext { get; set; } = null!;

    public TestContext VerifyTestContext { get; set; } = null!;
}
```

with MSTest 3.9.3 (which is the latest I see) `VerifyTestContext` is null.

MattKotsenas:
One potential fix is to mark [this ](https://github.com/microsoft/playwright-dotnet/blob/a878a633f888f3c920d4f057a9afe12368c036cf/src/Playwright.MSTest/WorkerAwareTest.cs#L13) as virtual in Playwright. That seems safe to do, but I'll continue to evaluate other potential options.

msexxeta:
> So it looks like the answer to both 1 and 2 is "no". When I use this simple repro:
[...]

PR #1489, opened 2025-07-27
Correctly handle the existence of `TestContext` in base class

Youssef1313:
Fixes https://github.com/VerifyTests/Verify/issues/1482

### DIFF
--- a/src/Verify.MSTest.SourceGenerator.Tests/InheritanceTests.HasAttributeOnDerivedClassAndPropertyManuallyDefinedInBase.verified.txt
+++ b/src/Verify.MSTest.SourceGenerator.Tests/InheritanceTests.HasAttributeOnDerivedClassAndPropertyManuallyDefinedInBase.verified.txt
@@ -12,7 +12,7 @@
 partial class Derived
 {
   [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Verify.MSTest.SourceGenerator", "1.0.0.0")]
-  public sealed override global::Microsoft.VisualStudio.TestTools.UnitTesting.TestContext TestContext
+  public new global::Microsoft.VisualStudio.TestTools.UnitTesting.TestContext TestContext
   {
     get => base.TestContext;
     set
--- a/src/Verify.MSTest.SourceGenerator.Tests/InheritanceTests.cs
+++ b/src/Verify.MSTest.SourceGenerator.Tests/InheritanceTests.cs
@@ -76,7 +76,7 @@ public class InheritanceTests(ITestOutputHelper output) : TestBase(output)
             }
             """;
 
-        return VerifyGenerator(TestDriver.Run(source), ["CS0506"]);
+        return VerifyGenerator(TestDriver.Run(source));
     }
 
     [Fact]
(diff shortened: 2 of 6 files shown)


MattKotsenas:
I don't think this is a safe change to do?

It suppresses the compiler warning, but now there may be two different properties, one that has the necessary machinery to hook the setter, and one that doesn't.

I think this could lead to Verify not seeing the correct TestContext if anything we're to interact with the base class?

Youssef1313:
@MattKotsenas There are two different properties, yes. One in base, and we shadow it in source-generated code generated by Verify. While shadowing is generally discouraged, I don't see much of an issue in that in this specific case.

The property generated by `Verify` is what MSTest will be dealing with, and the generated property is always setting the base one, so both properties are having the same value.

The only scenario where both properties can get out-of-sync is if something is setting the **base** property. But as `TestContext` property is only ever intended to be set by MSTest itself, there is not much of a concern in this case.

MattKotsenas:
I'll leave it to @SimonCropp if he wants to accept the change as project owner.

I'd still suggest against it, as though @Youssef1313 is correct about how things _should_ be, Verify will be at the mercy of all base classes behaving properly, and this change removes the only warning we currently have, which will make debugging these types of issues more difficult.

Instead I'd suggest either of the other two threads listed in the bug: fixing the Playwright tests as a targeted fix, or introducing the AsyncLocal to MSTest to sidestep the issue entirely.

However, if there's something I'm missing, please let me know!

Youssef1313:
For MSTest, I'm definitely interested in introducing a `Current` property. But I'm not sure if it's going to happen soon.

Another alternative for Playwright is to rely on constructor injection, but that forces all test classes to have a constructor injecting the TestContext to base.
Or, alternatively, MSTest could allow `TestContext` parameter to `TestCleanup` methods (that's where Playwright needs to access TestContext), but I'm not sure if it's a good idea to do so.

MattKotsenas:
@Youssef1313 , understood that the .Current property will likely take time. Isn't https://github.com/microsoft/playwright-dotnet/pull/3210 another alternative for Playwright that's non-breaking and allows this scenario?

SimonCropp:
@Youssef1313 

> For MSTest, I'm definitely interested in introducing a Current property. But I'm not sure if it's going to happen soon.

why? from my, perhaps naive, perspective, this should be fairly trivial thing to add?

could it perhaps be added faster as a `[Experimental]` feature?

Youssef1313:
> why? from my, perhaps naive, perspective, this should be fairly trivial thing to add?

Well, it's trivial thing to add and we already have the API internal. But we are already not very comfortable with `TestContext` API altogether, and we need to evaluate what our future vision for `TestContext` will be, and how adding such API will impact it. For example, we would like to start introducing new `TestContext` types at some point, where each `TestContext` type will be specific to a "stage" in the test lifecycle (e.g, assembly initialization, class initialization, test run, class cleanup, and assembly cleanup). The idea is that the TestContext of assembly initialization shouldn't have any properties that are irrelevant to assembly initialization. Today, `TestContext` passed to `AssemblyInitialize` has access to `TestName` (which is irrelevant to assembly initialization). Moreover, it doesn't throw today and it returns the name of the first test that will be run (in v4 we are changing it to throw).

So far, we don't yet have any concrete design for the specialized `TestContext` types we would like to introduce, and thus it's hard to know how much `TestContext.Current` API is going to have an impact. I'm **guessing** it's unlikely to have any big impact if we are going to leave the old `TestContext` for backcompat anyways, but still we need to be careful :)

Another thing is that I want to allow some time (and do more manual testing) around the propagation of `ExecutionContext`, which we have changed multiple times to fix some reported bugs.

SimonCropp:
closing this in fav of this https://github.com/microsoft/playwright-dotnet/pull/3210

Issue #1498, opened 2025-08-06
InvalidOperationException when serializing an object with a default `ImmutableArray`


johncrim:
#### Minimal Repro

```cs
    [Fact]
    public Task UninitializedImmutableArray() =>
        Verify(new HasImmutableArray());

    internal class HasImmutableArray
    {
        public ImmutableArray<int> Array;
    }
```

results in:

```
  Message: 
System.InvalidOperationException : This operation cannot be performed on a default instance of ImmutableArray<T>.  Consider initializing the array, or checking the ImmutableArray<T>.IsDefault property.

  Stack Trace: 
ImmutableArray`1.ThrowInvalidOperationIfNotInitialized()
ICollection.get_Count()
ReflectionHelpers.TryGetCollectionOrDictionary(Object target, Nullable`1& isEmpty, IEnumerable& enumerable) line 81
ReflectionHelpers.IsEmptyCollectionOrDictionary(Object target) line 66
SerializationSettings.TryGetScrubOrIgnoreByInstance(Object value, Nullable`1& scrubOrIgnore) line 42
CustomValueProvider.GetValue(Object target) line 23
JsonSerializerInternalWriter.CalculatePropertyValues(JsonWriter writer, Object value, JsonContainerContract contract, JsonProperty member, JsonProperty property, JsonContract& memberContract, Object& memberValue)
JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type type)
JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type type)
JsonSerializer.Serialize(JsonWriter jsonWriter, Object value)
VerifyJsonWriter.Serialize(Object value) line 401
Converter.Write(VerifyJsonWriter writer, InfoBuilder value) line 50
WriteOnlyJsonConverter`1.Write(VerifyJsonWriter writer, Object value) line 7
WriteOnlyJsonConverter.WriteJson(JsonWriter writer, Object value, JsonSerializer serializer) line 12
[...]

PR #1499, opened 2025-08-06
fix(reflection): Fix reflection case when collection is a default ImmutableArray<T>

johncrim:
This fixes an exception that occurs when serializing a type that has an `ImmutableArray<T>` property that hasn't been `new`ed.

Added tests for serialization and for ReflectionHelpers

fixes: #1498


johncrim:
@dotnet-policy-service agree company="Fintellect"

SimonCropp:
@johncrim thanks. i will deploy shortly

johncrim:
Thank you!

Issue #1497, opened 2025-08-05
$ref inside an array results in nulls in VerifyJson


saithis:
When I call VerifyJson method with JSON containing $ref tags inside an array, the result the tags are replaced with nulls.

```
[Fact]
public async Task Ref()
{
	var schema = """{"a":1,"b" : [ { "$ref" : "#/a" } ]},""";
	await VerifyJson(schema);
}
```

**Expected result**

```
{
  a: 1,
  b: [ 
    {
      $ref: #/a
    } 
  ]
}
```

**Actual result**

```
{
  a: 1,
  b: [ 
    null
  ]
}
```

It works for $refs that aren't inside of an array. That was reported here and is fixed now: https://github.com/VerifyTests/Verify/issues/1169
Unfortunatelly I couldn't find the exact commit that fixed it. It isn't linked and all the "refs and cleanup" commits are a red hering :(

SimonCropp:
@saithis happy to consider a pull request that fixes this

saithis:
@SimonCropp Could you point me to the commit that fixed it for #1169 ? That might help me find the issue faster. But I won't be able to work on this for at least the next 3 weeks. Afterwards I can try to find some time to take a shot at it.

saithis:
I was able to find some time today before I leave for my vacation. I hope the PR is fine.

SimonCropp:
yep. PR looks good. thanks

PR #1501, opened 2025-08-07
dont ignore json $refs inside of arrays

saithis:
This adresses #1497


saithis:
I don't understand why the PR check fails...

It shows:
```
Received: SerializationTests.Ref.received.json
{
  "a": {
    "$id": "x",
    "b": 1
  },
  "b": 2,
  "c": {
    "$ref": "#x/b"
  },
  "d": {
    "$ref": "#/b"
  },
  "e": [
    {
      "$ref": "#/b"
    }
  ]
}
Verified: SerializationTests.Ref.verified.json
{
  "a": {
    "$id": "x",
    "b": 1
  },
  "b": 2,
  "c": {
    "$ref": "#x/b"
  },
  "d": {
    "$ref": "#/b"
  }
}
```
So it seems like it didn't pick up my change to SerializationTests.Ref.verified.json and instead uses an old version of it?!?

SimonCropp:
> I don't understand why the PR check fails...

its because that test file is included in two projects. but i will fix that after merge

Thanks for this. i will deploy it now

PR #1552, opened 2025-09-26
Avoid unnecessary list instantiation.

fiseni:
Minor thing, you forgot to remove the instantiation.

PS. I also use this pattern extensively ([example](https://github.com/fiseni/Pozitron.Extensions.MediatR/blob/main/src/Pozitron.Extensions.MediatR/Publishers/WhenAllPublisher.cs)), and usually I inline it `(appendedFiles ??= []).Add()`.


SimonCropp:
good catch. thanks

PR #1579, opened 2025-10-20
Fix #1574 / F# Solution Explorer Reordering

### LINKED ISSUE #1574: Order of files/folders displayed in Visual Studio project explorer is wrong starting with Verify.NUnit 30.5 (F# project)
#### Describe the bug

I use `VerifyTest.NUnit` with Visual Studio 2022 latest version (17.14.16), inside an F# project.

Starting with `VerifyTest.NUnit` version 30.5, order of imported files in the project is messed up: a clean project displays fine, but as soon as `VerifyTest.NUnit` nuget package is resolved, order of files and folders displayed in the solution explorer changes.  
Folders containing `*.verified.txt` files are listed above everything else instead of staying in the normal order.

Image

---

I suspect this is somehow caused by the ability of `VerifyTests` to show `*.verified.txt` files as dependencies of tests, but it was working in versions before 30.5...

#### Minimal Repro

[WrongFolderOrderDemo.zip](https://github.com/user-attachments/files/22851833/WrongFolderOrderDemo.zip)

(See gif above demoing how to reproduce the issue with this solution)

Sorry I wasn't able to find the cause of the problem more precisely. I'm not sure what magic `VerifyTests` is doing to be able to show `*.verified.txt` files as dependencies of tests even when the files are not included explicitly.

BrunoJuchli:
## Description
Addresses items being reordered in Solution Explorer in F# projects as reported in #1574 by moving the `<None Include>` statements for verified/received files to Verify.targets - as was the case pre 30.5 / #1478 - for **F# only**.

## Tests
- Running `FSharpTests` passes
- Rider: In `FSharpTests`, *.verified.* do not show as nested. This is consistent with previous information. Rider shows the item metadata as expected, so.. still don't know why it isn't working correctly. As I've reported before, there seems to be an inconsistency with how Rider evaluates item metadata and the item metadata it shown in the dialog.
image

- I can't run tests in Verify.Tests locally:
  > MarkdownSnippets.MsBuild.targets(19,5): Error MSB4018 : The "MarkdownSnippets.DocoTask" task failed unexpectedly.
System.IO.FileNotFoundException: Could not load file or assembly 'System.Buffers, Version=4.0.5.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
File name: 'System.Buffers, Version=4.0.5.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'

### DIFF
--- a/src/DeterministicTests/DeterministicTests.csproj
+++ b/src/DeterministicTests/DeterministicTests.csproj
@@ -15,4 +15,5 @@
   </ItemGroup>
   <Import Project="$(ProjectDir)..\Verify\buildTransitive\Verify.props" />
   <Import Project="$(ProjectDir)..\Verify.Xunit\buildTransitive\Verify.Xunit.props" />
+  <Import Project="$(ProjectDir)..\Verify\buildTransitive\Verify.targets" />
 </Project>
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Condition="'$(ImportVerifyMsbuildFiles)' == 'true'"
+          Project="$(MSBuildThisFileDirectory)Verify\buildTransitive\Verify.targets" />
+</Project>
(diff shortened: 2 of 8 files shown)


BrunoJuchli:
@SimonCropp 
Could you produce a beta-version nuget package and then we ask @mlaily to test it?

SimonCropp:
@BrunoJuchli thanks so much for this

SimonCropp:
deploying this now as 31.0.4-beta.1

PR #1597, opened 2025-10-29
Use updated TUnit API

thomhurst:


SimonCropp:
@thomhurst thanks. this let me work out the changes in my own PR. I will deploy it now

PR #1600, opened 2025-10-31
process multiple member converters in case of inherited types

CommonGuy:


CommonGuy:
@dotnet-policy-service agree

PR #1614, opened 2025-11-07
TUnit 1.0.0

SimonCropp:


SimonCropp:
@thomhurst any ideas why this would start failing?

thomhurst:
I see they're around artifacts. I think I might have had some dodgy cache logic. Try v1.0.48 please and report back?

SimonCropp:
yep that works. fanks. will deploy now

PR #1708, opened 2026-04-07
Add Target comparer and avoid string allocations.

fiseni:


SimonCropp:
oh very nice. thanks

fiseni:
In addition, GroupBy is notorious allocation wise, and if willing to sort the list (or keep it sorted) it can be removed completely. I use this [approach](https://fiseni.com/posts/allocation-free-groupby/) in very hot paths.

PR #1729, opened 2026-05-10
Remove APIs obsolete since before v30

SimonCropp:


agross:
This broke the PdfPig 2.1.0 plugin that still uses this API in its latest release (2 years old!).

PR #1855, opened 2026-08-17
Format UTC offsets with invariant culture

SimonCropp:
`DateFormatter.GetDateOffset` rendered the offset numbers with plain interpolation, so they used `CurrentCulture` while every other format call in `DateFormatter` passes `Culture.InvariantCulture`.

Under a culture whose negative sign is not `-` (for example `ar-SA`, which uses U+061C followed by `-`), a `DateTimeOffset` parameter with a negative offset produced a parameter string, and hence a file name, that never matched a snapshot committed from an en-US machine. The same value also leaked into snapshot content via `Convert` when date scrubbing is off.

All four format calls now go through `Culture.InvariantCulture`. Output is unchanged for cultures that already use the invariant signs.

Added `DateFormatterTests.OffsetIsNotAffectedByCurrentCulture`, which formats under a culture with a custom negative sign. It fails on `main` and passes with this change. Full `Verify.Tests` suite passes.

### DIFF
--- a/docs/type-to-string-mapping.md
+++ b/docs/type-to-string-mapping.md
@@ -275,20 +275,20 @@ public static partial class DateFormatter
         {
             if (offset.Minutes == 0)
             {
-                return $"+{offset.TotalHours:0}";
+                return $"+{offset.TotalHours.ToString("0", Culture.InvariantCulture)}";
             }
 
-            return $"+{offset.Hours:0}-{offset.Minutes:00}";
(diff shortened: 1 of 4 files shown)


SimonCropp:
Duplicate of #1854, which covers the same fix with a wider theory (all five branches, including zero) and uses the real sv-SE U+2212 sign rather than an invented one. Closing in favour of that.

#1868 (sub-hour offsets) was stacked on this branch and is being rebased onto `fix-negative-offset-culture`.

PR #1868, opened 2026-08-17
Sign sub-hour UTC offsets

SimonCropp:
`GetDateOffset` took the sign from the hour component. For an offset of less than an hour that component is zero, so `TimeSpan.FromMinutes(-30)` rendered as `0-30` while its positive twin rendered `+0-30` — the two are indistinguishable by sign, and a negative offset silently reads as positive.

The sign now comes from the offset as a whole and the components are rendered unsigned. Every other case renders exactly as before: `-1-30`, `-5`, `+1-30`, `+5`, `+0`.

Only reachable with a constructed offset, since no real timezone sits in that range.

**Based on #1854**, not `main` — both rewrite the same four returns in `GetDateOffset`, so this is stacked to avoid a conflict. It keeps that PR's `FormattableString.Invariant` style and its theory test untouched; the culture pinning still matters for the digits, and the sign is now a literal rather than a formatted negative number, which the comment there has been updated to say.

`DateFormatterTests.SubHourOffsets` covers both signs and asserts the two no longer render the same. Full `Verify.Tests` passes on the rebased branch: 1305.

### DIFF
--- a/docs/type-to-string-mapping.md
+++ b/docs/type-to-string-mapping.md
@@ -268,39 +268,32 @@ public static partial class DateFormatter
     }
 
     // Interpolation formats with the current culture, and NumberFormatInfo.NegativeSign is not "-"
-    // everywhere: sv-SE renders U+2212 and ar-SA prefixes U+061C. A negative offset carries that
-    // sign into both snapshot content and parameter file names, so the culture is pinned here the
-    // same way it is for every date part above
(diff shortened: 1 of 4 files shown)


SimonCropp:
Superseded by #1872: GitHub closed this automatically when the base branch (#1854) was deleted on merge, and a closed PR cannot be reopened or retargeted once its base is gone. #1872 is the same commit rebased onto `main`.

PR #1910, opened 2026-09-03
Allow custom converters to override built-in converters

### LINKED ISSUE #1834: Converters from AddExtraSettings cannot override built-in converters
Converters added via `AddExtraSettings` currently land *behind* Verify's own converters, so they can never override a built-in.

`SerializationSettings.BuildSettings` adds ~25 built-in converters to `settings.Converters`, and only then runs the `extraSettings` callbacks:

```c#
converters.Add(dateConverter);
converters.Add(timeConverter);
// ...
foreach (var extraSetting in extraSettings)
{
    extraSetting(settings);
}
```

`AddExtraSettings` also applies the callback to the already-built `jsonSettings`, which likewise appends.

Argon resolves converters first-match-wins: `JsonSerializer.GetMatchingConverter` walks `Converters` by index and returns the first whose `CanConvert` returns true. So a user doing the natural thing:

```c#
VerifierSettings.AddExtraSettings(settings =>
    settings.Converters.Add(new MyDateOnlyConverter()));
```

gets a converter that is never invoked, with no error and no warning. Verify's `DateConverter` keeps winning. This came up in VerifyTests/Argon#394, where the user spent time debugging a `DateOnly` converter that could not have fired.

The workaround is `settings.Converters.Insert(0, ...)`, but that requires knowing both that Verify preseeds the collection and that Argon is first-match-wins. Neither is discoverable.

Suggestion: have Verify insert extra-settings converters ahead of its built-ins, so user converters take precedence by default. Options:

1. Run the `extraSettings` callbacks against the settings *before* adding the built-in converters, then append the built-ins. Simple, but changes ordering for any other converter-collection mutation the callback performs.
2. Snapshot `converters.Count` before running the callbacks and move newly added converters to the front afterwards. More targeted.
3. Leave the behaviour as is and document it, plus expose something explicit like `VerifierSettings.AddConverter` that inserts at index 0.

Option 1 or 2 seem preferable to me: a user-registered converter losing to a framework built-in is surprising in either direction, but silently losing is the worse of the two.

shubmittal:
## Summary

- Give converters added through `AddExtraSettings` priority over Verify's built-in converters.
- Preserve custom converter order when settings are changed or copied.
- Add tests for overriding, ordering, insertion, and removal.

Fixes #1834

## Validation

- 4 focused tests passed.
- `Verify.Tests` on .NET 11: 1,362 total, 0 failed.
- Release solution build passed.

### DIFF
--- a/src/Verify.Tests/Serialization/SerializationSettingsTests.cs
+++ b/src/Verify.Tests/Serialization/SerializationSettingsTests.cs
@@ -1,0 +1,91 @@
+public class SerializationSettingsTests
+{
+    [Fact]
+    public void ExtraConverterOverridesBuiltIn()
+    {
+        var converter = new CustomEnumConverter();
+        var settings = new SerializationSettings();
+        settings.AddExtraSettings(_ => _.Converters.Add(converter));
+
+        AssertUsesConverter(settings, converter);
+        AssertUsesConverter(new(settings), converter);
+    }
+
+    [Fact]
+    public void ExtraSettingsPreserveOrder()
+    {
+        var first = new CustomEnumConverter();
+        var second = new CustomEnumConverter();
+        var settings = new SerializationSettings();
+        settings.AddExtraSettings(
+            _ =>
+            {
+                _.Converters.Add(first);
+                _.ReferenceLoopHandling = ReferenceLoopHandling.Serialize;
+            });
+        settings.AddExtraSettings(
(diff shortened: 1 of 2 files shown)


shubmittal:
@dotnet-policy-service agree

SimonCropp:
thanks for this

